PR #33, opened 2019-05-09
Add a styles-mixin for sharing css sheets.

theengineear:
This sorta piqued my interest today, so I wanted to see what it might look like to try and follow the guidelines put out in https://wicg.github.io/construct-stylesheets/.

This PR raises some interesting questions:

1. Can we / should we attempt to make these styles non-modifiable? There's a glaring issue here in that you cache mutable objects... I couldn't figure out how to prevent modification.
2. Should the interface be simple strings? I like the simplicity here and non-forkiness of it.

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -14,6 +14,12 @@ custom extension is supported.
 Provides base functionality for creating custom elements with shadow roots and
 hooks to re-render the element.
 
+### `styles-mixin`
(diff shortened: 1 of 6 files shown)


theengineear:
Going to close this as a way to incorporate adopted stylesheets will be included as part of #58.

PR #37, opened 2019-05-11
Throw halting errors during analysis

theengineear:
This PR:

- updates the analysis phase to throw halting errors
- moves analysis into setup since we don't need to be connected to the DOM anymore.


theengineear:
Going to close this as static, one-time analysis with detailed error messages are provided as part of #58.

PR #48, opened 2019-10-09
Enforce eslint rules in travis.

theengineear:


etc/ready.js
```diff
@@ -1,14 +1,14 @@
 const ready = target => {
-  return new Promise((resolve, reject) => {
+  return new Promise(resolve => {
     if (target.readyState === 'complete') {
       resolve(target);
     } else {
-      function handle(evt) {
+      const handle = evt => {
```
theengineear: Fixes `no-inner-declarations`. I just switched it to an arrow function.

mixins/properties-mixin.js
```diff
@@ -28,6 +28,7 @@ export default superclass =>
       if (props) {
         return Object.keys(props).map(this.camelToDashCase);
       }
+      return undefined;
```
theengineear: Fixes `getter-return`.

mixins/property-effects-mixin.js
```diff
@@ -185,12 +185,15 @@ export default superclass =>
           return value ? '' : undefined;
         case 'String':
         case 'Number':
-          return value != null ? value.toString() : undefined;
-        default:
+          return value !== null && value !== undefined
```
theengineear: 🚨 Fixes `eqeqeq`. As far as I know, this was just looking to see if you were `null` or `undefined`. Otherwise, it was assuming you were property typed and `.toString()`-ing you. I think it's just more explicit now.

package.json
```diff
@@ -13,7 +13,7 @@
   "scripts": {
     "start": "element-server",
     "lint": "eslint --quiet .",
-    "lint-strict": "eslint .",
+    "lint-strict": "eslint --max-warnings=0 .",
```
theengineear: @klebba TIL `--max-warnings=0` to make sure that eslint fails on warnings.

test/test-scratch.js
```diff
@@ -92,6 +92,7 @@ it('scratch', async () => {
     el.shadowRoot.querySelector('span').textContent === '',
     'should update the DOM bindings'
   );
+  // eslint-disable-next-line require-atomic-updates
```
theengineear: I'll be honest, I don't fully understand why eslint was complaining here. However, it's in test code and I'm not too worried about adding this line.

theengineear:
@klebba , trying to tidy things up :)

mixins/property-effects-mixin.js
klebba: IMO we should rewrite this to be `return value === null || value === undefined ? undefined : value.toString()`

test/fixture-element-basic.js
```diff
@@ -25,7 +25,7 @@ class TestElement extends XElementBasic {
   }
 
   set booleanProperty(value) {
-    if (Boolean(value)) {
+    if (value) {
```
klebba: can't remember why I did this... but should be no change in behavior

mixins/property-effects-mixin.js
theengineear: Sure. I like that more 👍

test/test-scratch.js
klebba: is it a false positive? https://github.com/eslint/eslint/issues/11899

theengineear: Ah, great find. I'll add that to the comment here. Yes, I do think this is a false positive.

package.json
```diff
@@ -12,8 +12,8 @@
   },
   "scripts": {
     "start": "element-server",
-    "lint": "eslint --quiet .",
-    "lint-strict": "eslint .",
+    "lint": "eslint --max-warnings=0 .",
+    "lint-fix": "eslint --fix .",
```
theengineear: FYI @klebba , I think these are the two linting scripts we'll want moving forward.

theengineear:
Thanks!!

PR #49, opened 2019-10-14
allow creating custom directives

codeStryke:
https://lit-html.polymer-project.org/guide/creating-directives

Also refer to https://github.com/Polymer/lit-html/issues/877 for how directive can be used to manage uncontrolled input elements


klebba:
Nice!

PR #58, opened 2020-05-26
Simplify x-element.

theengineear:
Concept branch — the goal here is to reimagine `x-element` based on our learnings from using it in hundreds of components in a complex SPA over the last year.

Significant changes:

* Static analysis is done up-front, once.
* Anything that can be validated statically throws errors _except for interface shadowing_ which changes as browsers update — the best we can do is `console.warn` in this case.
* _Almost all_ of the XElement interface is static (though we provide trivial convenience wrappers). Note that `render` is the only non-trivial instance method.
* Static interface is ready for private (`#`) static fields when they're ready.
* `observer` >> `observe` is now a function (proper context is maintained if it's a constructor method).
* `computed` >> `input` + `compute`. `compute` is now a function (proper context is maintained if it's a constructor method). `input` is an array of property names which will be resolved to arguments at runtime and passed to `compute`.
* `value` is now `initial` and `default`. `initial` is used if initial value is nullish. `default` is used if property value is ever nullish.
* Type coercion is no longer done on set.
* Errors are thrown when you set a value with an incorrect type in the runtime.
* Concept of `readOnly` is first-class.
* Concept of `internal` is first-class.
* Internal setting/getting and read-only setting can be done via the `.internal` proxy.
* Computed properties are now lazily evaluated.
* Observation, reflection, and computation are no longer synchronous. Importantly, this allows us to set properties one-by-one and then get a single `compute` at the end.
* Concept of a "render root" is now supported via `createRenderRoot`.
* Initial template function is now cached on each host (vs defining a new function on each render).
* We always allow computation of computed properties before any arguments are defined. Previously, we had thought we might want to defer initial compute until at least one argument was defined.
* Swapped the order of values in `observe` callbacks to be `value, oldValue` — it was previously `oldValue, value`.
* Order of operations are now compute, reflect, render, then observe.
* Listener logic has been improved and interface is now a function, not a string (proper context is maintained if it's a constructor method).
* Coercion of `null` and `undefined` to `''` is removed, the underlying templating engine takes care of this for us.
* The notion of a render proxy is gone. Previously we had only used it to convert `null` and `undefined` to the empty string. However, that is handled by the templating engine now.
* Deserialization of anything other than `[Boolean, String, Number]` now throws a halting error (e.g., setting an attribute for a property that is of `type: Object`). Previously, we basically did nothing, especially for custom classes.


theengineear:
@klebba — mind taking a look when you have the chance? Also happy to walk you through the changes. I think you'll like the general direction here.

theengineear:
The most recent change is because I was using some functionality not supported in eslint yet because it's stage 3 functionality. I didn't think the code suffered much to walk it back, so I just made sure it adhered to stage 4 functionality.

theengineear:
@klebba — I left a couple things in limbo here. Please hold off on looking at the code until I get the chance to fix.

klebba:
I tried swapping this in for a feature I'm working and might have caught an issue; my observer functions are not being. Here's my property block:
```
  static get properties() {
    return {
      targetRect: {
        type: DOMRect,
      },
      open: {
        type: Boolean,
        reflect: true,
        observer: this.observeOpen,
      },
      _parentElement: {
        type: HTMLElement,
      },
      _cssProperties: {
        type: Object,
        dependencies: ['_parentElement', 'targetRect', 'open'],
        resolver: this.computeCSSProperties,
        observer: this.observeCSSProperties,
      },
    };
  }
```

My demo page has markup like:

```
<x-inspector debug open .targetRect="${rect1}">[...]</x-inspector>
```

So some of the values need to be read immediately following element upgrade. Also I just grabbed whatever I saw in the PR for `x-element.js` a few minutes ago so maybe it's not ready for this yet, but I thought I'd give it a try because I think the change in order for render/reflect etc might be helpful for this feature

theengineear:
>So some of the values need to be read immediately following element upgrade.

The initial update is done synchronously upon the first `connectedCallback`. This will cause, _synchronously_, your properties to be initialized, computed, reflected, rendered, and then observed.

I think that your component will be supported, the PR has just been thrashing a bit. Importantly, I think that properties block is looking good! As a reminder, the new way you'd want to write this would be like:

```javascript
class CustomElement extends XElement {
  static get properties() {
    return {
      targetRect: {
        type: DOMRect,
      },
      open: {
        type: Boolean,
        reflect: true,
        observer: this.observeOpen,
      },
      parentElement: {
        type: HTMLElement,
        internal: true,
      },
      cssProperties: {
        type: Object,
        internal: true,
        dependencies: ['parentElement', 'targetRect', 'open'],
        resolver: this.computeCSSProperties,
        observer: this.observeCSSProperties,
      },
    };
  }
}
```

theengineear:
@klebba I circled back on the issue you were having and perhaps there is a different bug that I'm misunderstanding. I've had a test in place the entire time to ensure that resolved, observed properties are synchronously called upon initial connection.

Would you mind trying to create a minimal example element that demonstrates your issue? It'd be super helpful!

theengineear:
OK, I'm willing to say that it makes sense to start trying to use this in your work as an experiment. I'm quite sure there will be some issues that need ironing out, but things should be stabilizing at this point.

klebba:
Let's start curating a migration guide:

was: 

```
static get properties {
  return {
    myProperty: {
       type: String,
       computed: 'computeMyProperty(foo, bar)',
    }
  }
}
```
now:

```
static get properties {
  return {
    myProperty: {
       type: String,
       resolver: this.computeMyProperty,
       dependencies: ['foo', 'bar'],
    }
  }
}
```
was:

```
this.listen(target, 'eventName',  callbackPointer);
this.unlisten(target, 'eventName',  callbackPointer);
```

now:

```
this.addListener(target, 'eventName',  callbackPointer, options);
this.removeListener(target, 'eventName',  callbackPointer);
```

klebba:
```
static async observePage(host, current, last) {
```

This throw an error:
```
Uncaught Error: Unexpected value for "XPages.properties.page.observer" (expected Function).
```

If I remove the `async` it appears to work

klebba:
`Uncaught Error: Unexpected value for "XPages.properties.currentElement" (expected HTMLElement).`

Would be nice to say `got ${value}` here as well

klebba:
```
Error: Property "XPages.properties.currentElement" has resolver (cannot be set).
```

Might read better as:

```
Error: Property "XPages.properties.currentElement" is read only. (defined by resolver ${resolverName}).
```

Or similar

klebba:
```
observer: this.observeModel,
```

Does not throw even though this function does not exist

klebba:
Here's an issue that seems like a bug:

```
    model: {
        type: Object,
        observer: this.observeModel,
      },
      currentElement: {
        observer: this.observerCurrentElement,
      },
      binding: {
        type: Boolean,
        internal: true,
        dependencies: ['currentElement', 'model'],
        resolver: this.computeBinding,
      },
```

The resolver function for `binding` is never called even though I'm able to confirm that both observers (`this.observerModel` and `this.observerCurrentElement`) are being called with the values I would expect. Note: I'm trying to port `x-pages`

theengineear:
>Also a question; if I throw in a resolver function what happens?

It will be a runtime error that will just keep reoccurring every time something asks for that value.

theengineear:
>If I remove the async it appears to work

Thanks, need to check for either `AsyncFunction` or `Function`.

theengineear:
>observer: this.observeModel,
>Does not throw even though this function does not exist

Yah, I've been loose about `undefined`. I'll change it to a `Reflect.has` check instead of `!== undefined`.

theengineear:
>The resolver function for binding is never called even though I'm able to confirm that both observers (this.observerModel and this.observerCurrentElement) are being called with the values I would expect. Note: I'm trying to port x-pages

Are you sure it's `this.computeBinding,` and not `this.resolveBinding`? I'm wondering if the loose `undefined` check was just problematic.

theengineear:
@klebba — not sure if I'd consider it a bug, but we specifically decided to move observation _after_ rendering. So I think that might be where the trouble is coming from. What if we just untangled the observer callbacks so you don't have observers setting values which cause observers?

```javascript
class XPages extends XElement {
  static get properties() {
    return {
      current: {
        type: String,
        reflect: true,
      },
      items: {
        type: Object,
      },
      model: {
        type: Object,
      },
      update: {
        type: Object,
        internal: true,
        dependencies: ['current', 'items', 'model'],
        resolver: (current, items, model) => ({ current, items, model }),
        observer: this.observeUpdate,
      },
    };
  }

  static observeUpdate(host, { current, items, model }, oldValue) {
    const container = host.shadowRoot.getElementById('container');
    if (current && current === oldValue.current) {
      // same page, re-bind the model
      container.firstChild.model = model;
    } else if (current && current !== oldValue.current) {
      // new page, clean up old page and create next page
      container.firstChild && container.firstChild.remove();
      const element = document.createElement(items[current]);
      container.appendChild(Object.assign(element, { model }));
    } else if (!current) {
      // no page, insert a slot so default content can be displayed
      container.firstChild && container.firstChild.remove();
      container.appendChild(document.createElement('slot'));
    }
  }

  static template(html) {
    return () => html`
      <style>
        :host,
        #container,
        #container > * {
          display: flex;
          flex-direction: column;
          flex-grow: 1;
        }
      </style>
      <div id="container"></div>
    `;
  }
}
```

theengineear:
>Would be nice to say `got ${value}` here as well

Yah, I haven't spent much time thinking about the messaging. I figured I'd start by at least making sure the things throw when we want them to throw.

klebba:
Thanks for all the fixes! Your x-pages refactor makes sense to me, but I still wonder why my callback never got called — maybe I'm missing something. Let's review together and I'll explain why I expected it to work

theengineear:
>but I still wonder why my callback never got called

Was it because you were using a resolver for a side effect and never looking at the actual value? Resolvers are lazily invoked, if you never ask for the return value, it never gets called. Moreover, they're memoized, we shouldn't depend on side effects in them.

theengineear:
@klebba — that code coverage though.

`# Coverage goal of 100% for http://0.0.0.0:8080/node_modules/@netflix/x-element/x-element.js met (got 100.00%).`

I actually caught a couple REAL BUGS attaining that 100% code coverage mark! Totally worth the effort 👍

klebba:
Wow, awesome!

klebba:
What do you think about adding a `lazy: false` flag to force evaluation of a computed property?

klebba:
The message `Property "DemoView.properties.guardRemove" is internal (element authors may use "internal")` might be more clear as `Property "DemoView.properties.guardRemove" is internal (internal.${prop})`

klebba:
I hit an issue where my dependency set for a computed property includes an internal property; `Property "XInspector.properties.myProp" is internal (element authors may use "internal").` — seems like this should work right?

theengineear:
>I hit an issue where my dependency set for a computed property includes an internal property; Property "XInspector.properties.myProp" is internal (element authors may use "internal"). — seems like this should work right?

It's hard for me to debug if I don't have an example. It shouldn't be an issue to include internal properties as dependencies for resolved properties. If you are seeing this error, then I've made a mistake for sure and I'll fix it.

I tried to repro, but I wasn't able to hit the error you're seeing.

theengineear:
>The message Property "DemoView.properties.guardRemove" is internal (element authors may use "internal") might be more clear as Property "DemoView.properties.guardRemove" is internal (internal.${prop})

Yah, I can reword. I'm just nervous about obscuring the fact that this is _for authors only_.

theengineear:
>What do you think about adding a lazy: false flag to force evaluation of a computed property?

Can you provide an example of when you'd want this?

klebba:
I figured out the issue with my `internal` pipeline; turns out it had nothing to do with the `internal` setting. The error message was a bit of a red herring. I had a name collision with a true DOM property — `parentElement` — and XElement freaked as a result. Not certain what we can do here; would be a little annoying to keep a list of know properties that could collide with user defined properties. Any thoughts?

theengineear:
Yep, we already keep a list of library-reserved properties. Try setting a property to `internal`. You should get a nice error.

As for shadowing in general, can we just lookup the property name after we've deleted it off the host during upgrade? I.e., if your prototype chain has that property already, throw an error.

klebba:
Minor comment:
```
throw new Error(`Property "${path}" is resolved (resolved properties are read-only).`);
```

Would be more clear as:
```
throw new Error(`Property "${path}" is computed (computed properties are read-only).`);
```

theengineear:
> `throw new Error(`Property "${path}" is computed (computed properties are read-only).`);`

Hmm, didn't we want walk away from the language of "computed"? This would be the only place that we'd use that word.

klebba:
My thinking here is that we still have computed properties, but the way we define them is to specify their dependencies and resolver function — a property definition which has these fields is considered computed — we could also call it `derived` if you want to disambiguate from old and new implementations

theengineear:
Got it. Yah, instead of me updating this right now, maybe we can bike shed as a group. Ideally, the language used in the property definitions matches the language we see in the error messages and matches naming conventions in the code.

klebba:
Sounds like a plan. I have an element that hooks into `afterInitialRender` -- do we still intend to include hooks like this?

klebba:
Total nit: I recommend we rename `propertiesProxy` to `propertyProxy`

klebba:
I have an observer that I expect to be called once; when the `newValue` is defined and the `oldvalue` is undefined I kick off a subroutine. However the observer function is called many times in succession with the same values for `newValue` and `oldValue`, making it difficult to detect the edge where `newValue` is first populated

theengineear:
>Total nit: I recommend we rename propertiesProxy to propertyProxy

🤔 but it's a proxy for the properties object. that doesn't make sense to me, but I'll change it if you want 👍

theengineear:
>Related: it seems like the render() method is no longer a public method available for override, breaking some of my integrations where I insert my code into this part of the element lifecycle

We can add it back. I'm just trying to make sure we _need_ all of these things. Can you paste an example of this to help me understand? I've never seen this method used outside of a `render` guard, which is an anti-pattern.

theengineear:
>I have an element that hooks into afterInitialRender -- do we still intend to include hooks like this?

In previous conversations, we decided that many of these hooks bloated our codebase. Let's discuss your use-case and decide if it's strictly necessary. I would really like to error on the side of having a smaller interface wherever possible.

theengineear:
>I have an observer that I use when a value is first defined; e.g. when newValue is defined and oldvalue is undefined I kick off a subroutine. However the observer function is called many times in succession with the same values for newValue and oldValue, making it difficult to detect the edge where newValue is first populated. This seems to differ from the last implementation, is this expected?

I must have introduced a bug. I thought I had a test for exactly this case, but I'll dig in tonight. You should never get called with the same arguments in an observer. Sorry about this!

theengineear:
@klebba , how would you feel about having the following:

```
renderCallback(initial) {
  console.log('this is the first time we rendered', initial);
}
```

A couple notes:

1. Note that there's no _skipping_ a render anymore. Previously, you could just not call `super.render` and that could potentially cause all sorts of confusion since the DOM won't have been created which can be super confusing. Instead, you are forced to render _something_, even if it is _nothing_.
2. Instead of two callback functions, we just give you a boolean flag `initial` which indicates whether or not you've rendered _for the first time ever_.
3. You cannot `render`. The rationale for this is that it can _again_ be a serious source of confusion. E.g., what happens if you manually call render before you've connected? The host won't have been initialized and it would be super confusing.
4. If we _must_ have a way to cause a render, I'd rather expose `invalidate` again, which would simply enqueue a render in the future under-the-hood.

At any rate, I _think_ the `renderCallback` (so named to match the conventions and flavors of other, native callbacks) may be enough for your needs.

theengineear:
>I must have introduced a bug. I thought I had a test for exactly this case, but I'll dig in tonight. You should never get called with the same arguments in an observer. Sorry about this!

I have a test case that specifically touches this functionality. Can you paste an example @klebba?

theengineear:
>Total nit: I recommend we rename propertiesProxy to propertyProxy

So, I did a bit of renaming such that internally, we have `propertyMap` and `attributeMap` and `listenerMap`. This allows us to deconflict referring to these objects as just `properties` and `internal` in our instance-side code. It's a bit of overloading, but I think that it ends up reading well.

In other words, `propertiesProxy >> properties` and `internalPropertiesProxy >> internal`.

theengineear:
>Related: it seems like the render() method is no longer a public

I did some refactoring here to to allow this again. As we discussed, you can break things if you misuse render right now (e.g., infinite loop because you set a property inside render). (I may still add a guard to prevent render from being called before the host is initialized).

klebba:
Working through more integration; I wonder if it would be a good idea to make it possible to return the last references from a computed function. The use case is that I want to scan the incoming dependencies and detect if I should mint a new reference (e.g. another array) or just return the last one. Since the reference won't have changed, downstream dependency resolvers would not be called

theengineear:
>I wonder if it would be a good idea to make it possible to return the last references from a computed function

That's an interesting idea and could be pretty neat:

```
static get properties() {
  return {
    a: {},
    b: {},
    c: { dependencies: ['a', 'b'], resolver: this.resolveC },
  }
}

// Just pass the last value.
static resolveC(a, b, oldValue) {/* .. */}

// Pass everything as an object.
static resolveC(a, b, { args: [lastA, lastB], value }) {/* .. */}
```

test/test-listeners.js
```diff
@@ -1,8 +1,84 @@
...
+  }
+  connectedCallback() {
+    super.connectedCallback();
+    this.listen(this.shadowRoot, 'custom-event', this.onCustomEventOnce, { once: true });
+    this.listen(this.shadowRoot, 'custom-event', this.onCustomEvent);
+    this.listen(this.shadowRoot, 'custom-event', this.constructor.onCustomEventStatic);
+  }
+  disconnectedCallback() {
+    super.disconnectedCallback();
+    this.unlisten(this.shadowRoot, 'custom-event', this.onCustomEventOnce, { once: true });
```
klebba, line 73: nit: not sure it's necessary to add `{ once: true }` here

theengineear:
@charricknflx @codeStryke @erahhal — Here's that PR I was talking about. I think a good starting point might be to read through the top-level comment (description) in this PR. Another good kickoff point would be to peruse the `SPEC.md` file in this change set. Obviously, feel free to dig into whatever you want though! Perhaps we can kick off a discussion more fomally on the topic next week.

klebba:
Caught an issue tonight; we can't use `BigInt` yet unless we want to drop support for Safari. I recommend we hold off on including it in the serializable types for now

theengineear:
@klebba 

* `BigInt` was removed to support Safari
* `XElement.createStyleSheet` was removed since it's not supported in all modern browsers
* I swept all the tests so that they all pass in all modern browsers
* I went through the tests and tried to add some commentary and clean things up in hopes that we'll be able to actually review them as a spec.
* I added an example of creating interstitial property bags to the spec.md file (i'll update the link in the comment)
* I found an esoteric bug in FF where somehow an element can be un-upgraded after being adopted into another document. I'm not going to bother to dig in right now and am conditionally skipping that test (required for 100% coverage) if the UA indicates that we're being run by Firefox.

klebba:
@theengineear great, thanks! I just pulled down the latest and my integrations are now working in Safari & Firefox — nice!

klebba:
Just noticed the type check in `__typeIsWrong` via `__setPropertyValue` and wondered if we should be concerned about the potential performance implications of checking types for all assignment operations. I came across this article which outlines our exact case: https://ponyfoo.com/articles/investigating-performance-object-prototype-to-string-es2015

Can you remind me why we do this type checking now?

theengineear:
>Can you remind me why we do this type checking now?

1. We used to do type coercion (which included type checking) on set, so we already had precedence for doing non-trivial work on set related to "type".
2. I didn't have this at first, but the team convinced me that we _should_ have it in a meeting on the topic.
3. It provides useful feedback to the integrator to improve development experience.
4. Without it, `type` would only be used as it relates to serialization and deserialization, so it falls back to being just an _annotation_ of the property, not necessarily doing anything. Also, it would be confusing to introspect a property value and get back an object when the property declares itself as a string.

My view on this is to do the thing that's helpful to development and to not prematurely optimize. If I'm understanding that article correctly it states that:

* This exact type checking is done in libraries as major as Angular and lodash
* The was able to get 6.5x improvement by avoiding general lookups on the `toStringTag` symbol by making changes to V8

Also, I would be shocked if _this_ was the thing that ultimately was the bottleneck in the update lifecycle of an x-element instance. I think maybe the actual ask here (if I can read between the lines) is that there is some fear, uncertainty, and doubt surrounding this change set that should be assuaged via perf testing before we can move forward.

It should be simple enough to do that. I can imagine that we would want to perf-test the following:

1. Element definition (i.e., what happens after we call `customElements.define`)
2. Element instantiation
  a. With computed properties
  b. With reflected properties
  c. With observed properties
3. Element update
  a. With computed properties
  b. With reflected properties
  c. With observed properties

I'll try to find some time to setup some sort of benchmarking. I don't have too much experience maintaining something like this though. For example, do you set a threshold or something and fail a test based on a delta from some baseline? I'm sure this is documented somewhere and I can look into it.

theengineear:
**CAVEAT** these tests are super nascent and we should not draw any final conclusions just yet!

A _really_ quick, cursory study of performance yields the following:

```
[THIS BRANCH]
average definition time (count = 100000): 0.028538199957692995 ms
average instantiation time (count = 100000): 0.006306099989451468 ms
average assignment time (count = 100000): 0.000526649983949028 ms
average update time (count = 100000): 0.0028123500145738945 ms

[MASTER]
average definition time (count = 100000): 0.013228350005229005 ms
average instantiation time (count = 100000): 0.01043335000926163 ms
average assignment time (count = 100000): 0.0005833499878644943 ms
average update time (count = 100000): 0.0028958999877795578 ms
```

I decided not to share the code since this is just an initial glance — we'd need more nuanced tests to really dig in. However, here's a quick analysis of the findings:

* As expected, definition time analysis takes longer now, ~2x slower to define an element. As a gut check, if we were to import and define _one thousand_ elements, it would still be just a fraction of our overall application boot time.
* Pertaining to the original thread — assignment and update remain nearly unchanged. I.e., there seems to be no bottleneck around type checking.
* As hoped for, there seems to be a discernible improvement in instantiation time — ~1.5 faster to instantiate an element.

x-element.js
```diff
@@ -1,15 +1,648 @@
...
-  )
-);
+import { render, html, svg, directive } from '../../lit-html/lit-html.js';
+import { asyncAppend } from '../../lit-html/directives/async-append.js';
+import { asyncReplace } from '../../lit-html/directives/async-replace.js';
+import { cache } from '../../lit-html/directives/cache.js';
+import { guard } from '../../lit-html/directives/guard.js';
+import { ifDefined } from '../../lit-html/directives/if-defined.js';
+import { repeat } from '../../lit-html/directives/repeat.js';
+import { unsafeHTML } from '../../lit-html/directives/unsafe-html.js';
```
klebba, line 8: is `unsafeSVG` overkill? any other directives left out here?

theengineear: also — `classMap`, `live`, `style-map`, `template-content`

codeStryke: can we get `live` in, it's useful for controlled components like `input`

theengineear:
@klebba — re-type-checking. I updated the comment where we were talking about it to mention that it doesn't seem to make things faster. THAT SAID, we haven't actually scrutinized the type checking I'm doing. It's fairly hard to do _perfect_ type checking in JS, so we should make sure that what I've implemented is even up to our standards before we talk about how performant it is.

x-element.js
theengineear: Updated `lit-html` to get all the new goodies and added them all. Note that we should probably make a quick PR to master that updates `lit-html` to ensure interop before merging and using this one now.

klebba:
Can you remind me: in a computed property pipeline, do computed scalars which have not changed in value still trigger downstream resolvers?

theengineear:
>In a computed property pipeline, do computed scalars which have not changed in value still trigger downstream resolvers?

Nope. Due to how we memorize, that should not happen anymore. Are you seeing that behavior?

klebba:
Oh great — I did not actually try to check this myself :)

theengineear:
Found some time last night to iterate on comments I got this week:

* Return value of `createRenderRoot` is now validated and throws helpful errors.
* Improvements to error messaging — instead of printing the value, we now print the _type_ of the value, which is often more useful. At a minimum, it's much more consistent now.
* Autocomplete behavior in Chrome devtools' console now works when introspecting the `.internal` proxy during debugging (same goes for `properties`, though you probably won't be debugging `properties` via the proxy since you would just use the host).

And — after killing some braincells, I realized that we've historically been checking for method existence using `Function.name` (e.g., `object[function.name] === function`). However, (1) `Function.name` is mutable, (2) the name of the function need not match the name from the object, and (3) I believe code minification tools may cause issues as well. The only reason I found this is that through some perf testing, I realized that `Function.name` is surprisingly slow! My solution is to walk the constructor tree or the prototype tree and loop over `ownKeys`. Amazingly, that seems to be just as performant as the slow name lookup and it's more correct `</rant>`.

theengineear:
Note to self — make sure that we throw hard if an initial property or attribute value exists for a resolved property. Right now, I think it's just ignored, but we actually throw an error in the case.

theengineear:
>make sure that we throw hard if an initial property or attribute value exists for a resolved property

We _sort of_ do this. If you set the initial value of a resolved property to the value `undefined` _it will not throw_. I'm going to change this since I think it's actually incorrect. Our rule should be "you cannot initialize a resolved property" NOT "you cannot initialize a resolved property to anything but `undefined`".

TL;DR — we were validating initialization, but there's a subtle edge case I'm also going to guard against.

theengineear:
@erahhal and @klebba — I think I've talked to both of you about this one. How do you feel about the name `value` representing the initial, default value of a property? Can we force this to be a function and call it `initializer`?

```
{
  myProp: {
    type: String,
    readOnly: true,
    initializer: () => 'foo',
    observer: () => {},
  },
}
```

Some rationale:

1. `initializer` conveys the usage much better than `value`.
2. Matches verbage / language we use elsewhere.
3. There'd be one-and-only-one way to initialize — a function.
4. Static validation of `initializer` is much simpler than static validation of `value` (in its current form).

If we were being pedantic we'd call it `defaultInitializer`. However, I think the ergonomics of that are bad enough to rule it out.

theengineear:
Re: functional interface — here's a radical idea. What if we removed all the magical binding completely by forcing you to write anonymous functions. Additionally (back to a DSL) via string introspection, we can switch on the underlying behavior in a more declarative fashion.

```
static get properties() {
  return {
    a: {
      type: Number,
      initializer: () => 5,
    },
    b: {
      type: Number,
      initializer: () => 10,
    },
    c: {
      type: Number,
      resolver: this.resolveC,
      observer: this.observeC,
    },
  };
}

static get listeners() {
  return {
    click: this.onStaticClick,
  };
}

static resolveC = (a, b) => a + b;
// static resolveC = (accumulator, a, b) => a + b;

static observeC = (host, current, previous) => { /* .. */ };
// static observeC = (current, previous) => { /* .. */ };
// static observeC = (current) => { /* .. */ };
// static observeC = current => { /* .. */ };

static onStaticClick = (host, evt) => { /* .. */ };
// static onStaticClick = (evt) => { /* .. */ };
// static onStaticClick = evt => { /* .. */ };

onInstanceClick = evt => { /* .. */ };

connectedCallback() {
  this.shadowRoot.addEventListener('click', this.onInstanceClick);
}

disconnectedCallback() {
  this.shadowRoot.removeEventListener('click', this.onInstanceClick);
}
```

Note how the idiomatic binding removes our need for the `listen` and `unlisten` helpers.

Also note that we remove `dependencies` because we can do strict pattern matching on the `Function.toString()` — I think it'd be cool to try and favor `dependencies` if they're explicitly declared and fallback to `Function.toString()` by default.

** FYI the instance declarations are forthcoming in future versions of JS. That said, you can wrap in an anonymous function or `bind` in the constructor for static and instance methods today.

theengineear:
☝️ @klebba @erahhal @charricknflx @codeStryke — 🚨 some wild ideas above. Just wanted to float some musings I've had while working on this PR. I can't help but wonder whether anonymous functions declared as idiomatic fields via the stage 3 proposal could obviate the need for all the binding backflips we do in our code.

I opened up a specific ticket (https://github.com/tc39/proposal-class-fields/issues/312) related to this need on the proposal for class fields.

One cool thing about going this direction is that we have workarounds _today_ for this behavior:

1. Wrap your static method in an anonymous function in the properties block — done.
2. Call `.bind(this)` on your instance method in the constructor block — done.

I think we could delete a _lot_ of code if we required that users follow certain patterns.

klebba:
Will review these comments, but wanted to leave a note before I forget that your mouse-ghost appears to no longer work :(

theengineear:
Ha, I deleted it. In retrospect I felt that it was a little lame 😅

x-element.js
```diff
@@ -1,15 +1,721 @@
...
+      throw new Error(`Unexpected value for "${path}.attribute" (expected String, got ${typeName}).`);
+    }
+    if (Reflect.has(property, 'attribute') && attribute === '') {
+      throw new Error(`Unexpected value for "${path}.attribute" (expected non-empty String).`);
+    }
+    if (
+      Reflect.has(property, 'value') &&
+      value !== undefined &&
+      value !== null &&
+      XElement.__typeIsWrong(Boolean, value) &&
```
klebba, line 218: This check threw me for a loop; I was trying to assign `new URL('iframe/index.html', import.meta.url)` as my value without using a function, since I don't mind if instances share the URL object. The error message was confusing because I first thought it was referring to the URL type I had declared as the property type.

theengineear:
>The error message was confusing because I first thought it was referring to the URL type I had declared as the property type.

Yah the suffixing `.value` can be hard to miss. I'll be more informative in the error message. The nice thing is that error message updates are exceedingly simple to change now 👍

theengineear:
TODO:
  * test `initial`
    * ensure it be a constructor method
  * test `default`
    * ensure it be a constructor method
    * if initial is nullish, ensure we get back the default
    * test that computed values are type-checked.

klebba:
I hit an issue with a `compute` function that stopped working. This element presents the problem: https://gist.github.com/klebba/cd795a01223beaa219c8560db1ec512c — the `computeBinding` function is never actually called. I tried renaming and adding other functions with different inputs, but none were called. Any ideas?

klebba:
Is it possible to throw an error when a `listen` or `unlisten` callback is not found? Would be nice feedback

theengineear:
>I hit an issue with a `compute` function...

Yep, we stopped forcing the compute callbacks inside render completely now. I don't understand why setting a dummy `observe` wouldn't work though 🤔

```
static get properties() {
  return {
    a: {},
    b: {},
    c: {
      input: ['a', 'b'],
      compute: (a, b) => a.sideEffect(b),
      observe: () => {}, // dummy observer to force a sideEffect in compute.
    },
  };
}
```

☝️ I'd expect that to work. Note that observe is called _after render_ — could that be the issue? You're getting the side effect, but it's happening only render has occurred? If that's the case, you could also pluck it from the properties proxy inside your render routine.

```
static get properties() {
  return {
    a: {},
    b: {},
    c: {
      input: ['a', 'b'],
      compute: (a, b) => a.sideEffect(b),
    },
  };
}

static template(html) {
  return properties => {
    properties.c; // forces a side effect inside compute
    const { a, b } = properties;
    return html`<div>${a}</div><div>${b}</div>`;
  }
}
```

theengineear:
>Is it possible to throw an error when...

For sure, we do that for the static listeners block — seems consistent to do it when using `listen` / `unlisten` as well.

theengineear:
@klebba — `listen` and `unlisten` through insightful errors when passed something other than a `Function` now.

klebba:
Based on our conversation I reworked my `x-pages` variant here: https://gist.github.com/klebba/cd795a01223beaa219c8560db1ec512c#file-x-pages-js-L41 — (note line 41) — this cut down quite a bit on lines of code and should actually "do about the same thing"

theengineear:
@klebba — the `host` is now passed as the second argument to the template callback again.

theengineear:
@codeStryke — as requested, I updated the documentation in the source code.

klebba:
@theengineear can you remind me what all remains here? If my memory serves, we're really close! 🥇

theengineear:
We need to complete the refactor in our application to prove that we like
this change set. Then, we can merge / publish. There's no real reason to
merge this PR in before we've finished testing it on the ~500 components we
have IMO. At a minimum, I'd like to get 50% of these components refactored
before merging this in. We can always merge it and then wait to publish.
It's just that once we merge it in, we break our ability to introduce any
sort of hot-fix.

theengineear:
OK — new branch, new PR. I'm going to close down this exploratory PR in favor of a fresh start (just a copy of what's here).

Let's continue the discussion (if there's more to discuss) on #69.

PR #62, opened 2020-08-05
[POC] Documentation

codeStryke:
In the past I've found the annotated source code of [backbone.js](http://backbonejs.org/docs/backbone.html) and [underscore.js](http://underscorejs.org/docs/underscore.html) very useful.

This is an attempt to see if we can maintain useful documentation about x-element in a low cost way.

image


theengineear:
@codeStryke — figured I'd ping you to see if there's any updates on this. Should we (1) leave it as is (2) update (3) close?

PR #69, opened 2020-12-09
Next iteration of x-element (soon to be 1.0.0).

### LINKED ISSUE #25: Cache template function for use in render
It seems like we should cache the response of `static template` so that we could re-use it efficiently. Is there a reason we're not doing this?

### LINKED ISSUE #26: Switch order of observer arguments
Currently, you get the values in the following order: `oldValue, newValue`

This was done to match `attributeChangedCallback`. However, in practice this is not as ergonomic. It's often the case that you do not care about the previous value, causing code like this: `myObserver(_, value)`.

This is pulled into a separate issue since it will cause a sweeping change of all code leveraging observers.

### LINKED ISSUE #28: Cache static analysis per-class and reuse during initialization.
#23 improves the factoring of x-element. Another refactor we would like to take on is handling analysis at _first construction_ and caching the result to be shared across all future instances.

Currently, we do the same "static" analysis per instance, which is mostly duplicative. Ideally, we would be able to `initialize` with a cached `analysis` value which basically just binds the current instance to all the methods etc.

Related to #31

### LINKED ISSUE #30: Force observer and computed methods to be static
This is related to #28. Also, it's potentially a nice way for us to remove a fork that may not be providing much benefit.

Benefits to making these static:
- no chance of naming collision for instance level properties 
- one less fork in our code
- information can be used in static analysis since it doesn't depend on an instance

Downsides to making these static:
- a bit more restrictive
- observer callback may fell a bit less ergonomic since we'll change the call signature

For `computed`... There's not too much argument here, there should be no need to provide a `target`. Forcing these methods to be `static` will also help with potential, future memoization of computations.

For `observer`... We want to support access to the `target` still, so we would need to change the signature to be `myCallback(target, value, oldValue)`--or something like that.

### LINKED ISSUE #31: Analysis-time errors should be halting
Currently, analysis work actually uses an instance (to resolve methods, etc). This allows us to dispatch errors at runtime instead of just throwing errors. Because we're not throwing halting errors, we're able to consider these _errors_ non-fatal and move on.

In light of the comment below about the truly static `observeAttributes` hook and in light of discussions about making both computed methods and observers static #30, I propose we _do_ cause a halting error.

It will:

1. Remove forks in logic that deal with graceful failure
2. Allow us to be _truly_ static with our analysis
3. Can't be missed by developers since it will fail to render anything

As long as analysis is truly static, you know that if an instance boots in development, it will boot in production, so it's not really possible to cause an error in production if you got the thing working in development.

### LINKED ISSUE #36: Discussion: Should we get rid of the computed DSL?
The DSL was a convenience ported over from Polymer. However, this forces us to do some additional parsing work. From the perspective of staying lean, this could just be considered bloat.

Here's a potentially better, more declarative way forward:

```
computed: ['computeC', 'a', 'b']
```

That doesn't feel terribly hard to write out to me and then there's basically no guesswork as to what could be happening here.

### LINKED ISSUE #42: readOnly property initial value can't be set
```
static get properties() {
    return {
      readOnlyProperty: {
        type: String,
        value: 'foo',
        readOnly: true,
      },
    };
  }
```

I would expect this to work — Polymer had this behavior so that you could configure constants that could be fed through the property pipeline

### LINKED ISSUE #46: Expose lit-html's "svg" for creating svg template results.
See https://lit-html.polymer-project.org/api/modules/lit_html.html#svg

And here's the source: https://github.com/Polymer/lit-html/blob/master/src/lib/template-result.ts#L134-L144

Without `svg`, you cannot do the following:

```
static templateCircle(html, circle) {
  return html`
    <!-- This doesn't work, lit-html never sees an svg tag to set context. -->
    <circle cx="${circle.x}" cy="${circle.y}" r="${circle.r}"></circle>
  `;
}

static template(html) {
  return ({ circles }) => {
    return html`
      <svg xmlns="http://www.w3.org/2000/svg">
        <!-- This works, lit-html sees the svg tag and sets context. -->
        <circle cx="0" cy="0" r="10"></circle>
        ${circles.map(circle => this.renderCircle(html, circle))}
      </svg>
    `;
  };
}
```

The tricky part will be figuring out the correct interface here. Let's update this as we come up with new ideas:

```
static templateCircle(svg, circle) {
  return svg`
    <circle cx="${circle.x}" cy="${circle.y}" r="${circle.r}"></circle>
  `;
}

static template({ html, svg }) {
  return ({ circles }) => {
    return html`
      <svg xmlns="http://www.w3.org/2000/svg">
        <circle cx="0" cy="0" r="10"></circle>
        ${circles.map(circle => this.renderCircle(svg, circle))}
      </svg>
    `;
  };
}
```

### LINKED ISSUE #57: Remove use of eval()
Stub ticket.

### LINKED ISSUE #66: XElement should warn when improperly shadowing "data-*" attributes.
Similar to the `class` attribute, `data-*` attributes are meant for the _integrator_, not for the custom element author. When composing custom elements, it is expected that there won't be any conflict when you set `data-*` attribute on an element.

Therefore, `x-element` should WARN authors when they've inadvertently created a property name (e.g., `dataValue`) since that is associated with the attribute `data-value` — which has special handling (i.e., syncing to the `.dataset.value` property).

### LINKED ISSUE #67: XElement should warn authors when shadowing asymmetric inherited attributes.
While _many_ attributes have a simple relationship to properties (e.g., `.title` >> `"title"`), there isn't a clear standard for naming conventions when it comes to properties reflecting to attributes. Some interesting cases:

```
class <<>> .className
contenteditable <<>> .contentEditable
aria-valuemin <<>> .ariaValueMin
itemscope <<>> ???
```

XElement should attempt to WARN authors when they're inadvertently shadowing attribute interfaces based on mappings between attributes and property names.

* Here's a list of global attributes (for reference) https://html.spec.whatwg.org/multipage/dom.html#global-attributes
* We should also consider all `aria-*` attributes as special cases.

### LINKED ISSUE #68: Invalid template results can cause cryptic errors.
A template result can successfully be created and _then fail to render_. Because it's common that an element will only override the "template" method, the stack trace will not actually print the name of the problematic file — this makes debugging _incredibly_ difficult.

One approach would be to wrap the call to lit-html's `render` in a `try-catch` such that we can append DOM pathing information when we encounter an error.

Example cryptic stack trace:

```
Uncaught TypeError: items is not iterable
    at repeat.ts:122
    at NodePart.commit (parts.ts:210)
    at TemplateInstance.update (template-instance.ts:53)
    at NodePart.__commitTemplateResult (parts.ts:280)
    at NodePart.commit (parts.ts:221)
    at render (render.ts:55)
    at HTMLElement.render (x-element.js:158)
    at Function.__updateHost (x-element.js:838)
    at Function.__initializeHost (x-element.js:742)
    at HTMLElement.connectedCallback (x-element.js:131)
```

But we _could_ modify the message to be:

```
Uncaught TypeError: items is not iterable — Invalid template for "HelloElement" at path "hello-world < div#container"
    at repeat.ts:122
    at NodePart.commit (parts.ts:210)
    at TemplateInstance.update (template-instance.ts:53)
    at NodePart.__commitTemplateResult (parts.ts:280)
    at NodePart.commit (parts.ts:221)
    at render (render.ts:55)
    at HTMLElement.render (x-element.js:158)
    at Function.__updateHost (x-element.js:837)
    at Function.__initializeHost (x-element.js:741)
    at HTMLElement.connectedCallback (x-element.js:131)
```

In the later case, you _immediately_ know where to start debugging. You may not know why `items is not iterable`, but at least you know where to look for the culprit.

theengineear:
This closes #68, closes #67, closes #66, closes #57, closes #46, closes #42, closes #36, closes #31, closes #30, closes #28, closes #26, and closes #25.

This is a clean version of #58 — that PR began as an exploration and turned into a promising candidate for our initial 1.0.0. Since the discussions got lengthy in there, I decided to start fresh with a better branch name and a clean slate.

Significant changes:

* Static analysis is done up-front, once.
* Anything that can be validated statically throws errors
* Environment / runtime errors `console.warn` (e.g., inherited property shadowing, `data-*` attributes, etc.)
* _Almost all_ of the XElement interface is static.
* `observer` >> `observe` is now a function (proper context is maintained if it's a constructor method).
* `computed` >> `input` + `compute`. `compute` is now a function. `input` is an array of property names which will be resolved to arguments at runtime and passed to `compute`.
* `value` >> `initial` + `default`. `initial` is used if initial value is nullish. `default` is used if property value is ever nullish.
* Type coercion is no longer done on set.
* Errors are thrown when you set a value with an incorrect type in the runtime.
* Concept of `readOnly` is first-class.
* Concept of `internal` is first-class.
* Internal setting/getting and read-only setting can be done via the `.internal` proxy.
* Computed properties are now lazily evaluated.
* Observation, reflection, and computation are no longer synchronous. Importantly, this allows us to set properties one-by-one and then get a single `compute` at the end.
* Concept of a "render root" is now supported via `createRenderRoot`.
* We always allow computation of computed properties before any arguments are defined. Previously, we had questioned that approach.
* Swapped the order of values in `observe` callbacks to be `value, oldValue` — it was previously `oldValue, value`.
* Order of operations are now `compute`, `reflect`, `render`, then `observe`.
* Listener logic has been improved and interface is now a function, not a string.
* All callbacks are bound to the constructor (e.g., `compute`, `listen`, `observe`, etc).
* Coercion of `null` and `undefined` to `''` is removed, the underlying templating engine takes care of this for us.
* Deserialization of anything other than `[Boolean, String, Number]` now throws a halting error (e.g., setting an attribute for a property that is of `type: Object`).


theengineear:
FYI @klebba — I think we're finally ready to merge this!

theengineear:
Note — we will probably bump to `1.0.0-rc.41` before _finally_ moving to `1.0.0`. The important part is that we should be clear to merge this in and release 👌

klebba:
@theengineear awesome work!

PR #72, opened 2020-12-11
Shorten warning messages.

### LINKED ISSUE #71: Stack trace for warnings is unhelpful.
If you get a warning due to a validation issue, we create an error and then print it to the console. This isn't actually super helpful though:

```
Error: Unexpected key "HelloElement.properties.tabIndex" has attribute "tab-index" which is related to the inherited property "tabIndex", behavior not guaranteed.
    at Function.__validateProperties (x-element.js:307)
    at Function.__analyzeConstructor (x-element.js:254)
    at Function.get observedAttributes [as observedAttributes] (x-element.js:22)
    at index.js:84
```

The only important part is really the message there. We might as will shorten this a bit.

```
Unexpected key "HelloElement.properties.tabIndex" has attribute "tab-index" which is related to the inherited property "tabIndex", behavior not guaranteed.
```

theengineear:
We previously passed an Error into the console.warn call. However, the
stack trace isn’t actually helpful. Instead, we’ll just pass the error
message.

Closes #71

### DIFF
--- a/test/test-analysis-errors.js
+++ b/test/test-analysis-errors.js
@@ -62,8 +62,8 @@ it('properties should not shadow prototype chain interface', () => {
     'Unexpected key "TestElement.properties.title" has attribute "title" which is related to the inherited property "title", behavior not guaranteed.',
   ];
   const consoleWarn = console.warn; // eslint-disable-line no-console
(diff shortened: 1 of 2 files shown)


theengineear:
FYI @erahhal , @codeStryke , @charricknflx

PR #76, opened 2020-12-17
Relax access to internal properties.

### LINKED ISSUE #74: data-* warning irrelevant for internal properties
Given that `internal` properties may not be attribute bound or reflected, it probably makes sense to relax this warning in this case.

theengineear:
Previously, to help developers, we threw errors when trying to set/get
attributes/properties for “internal” properties. Instead of shouting
about internals, we’re going to take a new tact — silence.

Setting/getting property values for internals will now _do nothing_.
This has the excellent side-effect of freeing up internal properties to
be named _anything_ since they’re safely name-spaced inside
“host.internal”.

E.g., you can now have an internal property called “title”, which
doesn’t interfere _at all_ with the inherited property “title”, nor the
inherited attribute “title”.

Closes #74.

### DIFF
--- a/x-element.js
+++ b/x-element.js
@@ -20,8 +20,7 @@ export default class XElement extends HTMLElement {
   /** Extends HTMLElement.observedAttributes to handle the properties block. */
   static get observedAttributes() {
     XElement.__analyzeConstructor(this);
-    const { attributeMap } = XElement.__constructors.get(this);
-    return [...attributeMap.keys()];
+    return [...XElement.__constructors.get(this).attributeMap.keys()];
   }
 
   /**
@@ -267,7 +266,9 @@ export default class XElement extends HTMLElement {
         internalTarget[key] = undefined;
       }
       propertiesTarget[key] = undefined;
-      attributeMap.set(property.attribute, property);
+      if (property.attribute) {
+        attributeMap.set(property.attribute, property);
+      }
     }
     const listenerMap = new Map(listenersEntries);
     XElement.__constructors.set(constructor, {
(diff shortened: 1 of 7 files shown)


x-element.js
```diff
@@ -291,31 +293,34 @@ export default class XElement extends HTMLElement {
...
-      }
-      if (PROBLEMATIC_ATTRIBUTES.has(attribute)) {
-        const { xElementPropertyName, inheritedPropertyName } = PROBLEMATIC_ATTRIBUTES.get(attribute);
-        if (xElementPropertyName) {
-          throw new Error(`Unexpected key "${path}.${key}" has attribute "${attribute}" which is related to an x-element property "${xElementPropertyName}".`);
-        } else if (inheritedPropertyName) {
-          console.warn(`Unexpected key "${path}.${key}" has attribute "${attribute}" which is related to the inherited property "${inheritedPropertyName}", behavior not guaranteed.`); // eslint-disable-line no-console
-        } else {
-          console.warn(`Unexpected key "${path}.${key}" has attribute "${attribute}" which is inherited, behavior not guaranteed.`); // eslint-disable-line no-console
+      if (!property.internal) {
```
klebba, line 296: nit: bail early instead to cut down on the indent
```
if (property.internal) {
  return;
}
```
instead

x-element.js
```diff
@@ -748,25 +755,22 @@ export default class XElement extends HTMLElement {
...
-      : () => XElement.__getPropertyValue(host, property);
-    const set = compute || readOnly || internal
-      ? () => {
-        if (compute && !internal) {
-          throw new Error(`Property "${path}" is computed (computed properties are read-only).`);
-        } else if (readOnly) {
-          throw new Error(`Property "${path}" is read-only (internal.${key}).`);
-        } else {
-          throw new Error(`Property "${path}" is internal (internal.${key}).`);
+    if (!property.internal) {
```
klebba, line 758: same nit

x-element.js
```diff
@@ -937,33 +941,33 @@ XElement.__propertyKeys = new Set(['type', 'attribute', 'input', 'compute', 'obs
 XElement.__serializableTypes = new Set([Boolean, String, Number]);
 XElement.__caseMap = new Map();
 
-const PROBLEMATIC_PROPERTY_NAMES = new Map();
-const PROBLEMATIC_ATTRIBUTES = new Map();
+const BAD_PROPERTY_NAMES = new Map();
```
klebba, line 944: At a glance it's not clear what `bad` means. Is it synonymous with `RESERVED_PROPERTY_NAMES`?

klebba:
I did a quick pass, mostly for style — lots of deleted code here! I will admit that the developer errors when accessing internals have helped save me some time once or twice before, mostly when refactoring, but I don't know what I will greatly miss them

theengineear:
Ha, yes, I find myself feeling the same way. I made this comment earlier today in a conversation:

>I was super hesitant about doing this because I found it useful to have the framework tell me when I’m trying to access an internal property through the host — however, now that we’re through the migration, I think the costs actually outweigh the benefits.

x-element.js
theengineear: Just trying to cut PROBLEMATIC down. `INHERITED` was what I originally picked, but it's not quite right, the property `className` is inherited, but the property `class` is _problematic_ since it reflects to the _inherited_ attribute `class`.

Naming things is hard. I'll try to spare a few more brain cells when I'm back in the code.

x-element.js
theengineear: Hmm, I often prefer to leave the indent since it makes the code path easier to reason about — in this case using `continue` is akin to having a function with more than one `return` statement.

Also, I think in this case, that would cause a bug since the `if (property.input)` condition _can still be met_ even if the property is internal.

x-element.js
theengineear: I changed it to reserved so that there's no contention — I also updated our error messaging and the test checking that error messaging to match.

PR #81, opened 2021-02-23
Make functional “initial”/“default” per instance.

theengineear:
Previously, we allowed these to be shared across all instances, it’s now
only shared over time across the same instance.

Importantly, the same default value is used per-instance over and over
again. Otherwise, you might unnecessarily cause additional computes when
an upstream dependency gets a _new_ default reference.

IMPORTANTLY, this means that if you wrote code like `default: () => () => (\* whatever *\)`, you need to change it to `default: () => (\* whatever *\)`. Previously, that's how you would make something _per instance_ vs _per constructor_. I.e., whereas the functional syntax served as a reminder that you were setting something mutable, you now just _cannot_ set a mutable default on the constructor _at all_.

### DIFF
--- a/test/test-default-properties.js
+++ b/test/test-default-properties.js
@@ -26,18 +26,14 @@ class TestElementAnonymous extends XElement {
         default: (() => 'one')(),
       },
       two: {
-        type: String,
-        default: () => 'two',
-      },
-      three: {
-        type: String,
-        default: () => () => 'three',
+        type: Object,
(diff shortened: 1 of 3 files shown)


theengineear:
@codeStryke and @klebba — I updated this to match our current expectations. The test suite was update (and extended) to match.

Gonna merge since I doubt there will be any objection here. I made a note about the lines of code that might potentially be effected. They're easy to search for in a codebase.

Issue #80, opened 2021-02-19
change the default behavior to not reflect properties as attributes


codeStryke:
opening up an issue to discuss changing the default behavior to not automatically reflect every property as an attribute. Instead, we can rely on the `attribute` field of a `property` that we want observed. This change will makes the interface more explicit, avoid shadowing attributes like `id` or `data-*`.
@theengineear @klebba

klebba:
@codeStryke only scalar values may be reflected as attributes; boolean, number, string — and the behavior is not automatic. The developer must opt in using `reflect: true` in their property definition.

Our philosophy on overloading attribute names is that the library should not get in the way of the developer. The developer is thus expected to understand the way the browser interacts with attribute names they choose.

For example what if I actually do want to bind `id` or `tabindex` or `title` — all are native [global attributes](https://www.w3schools.com/tags/ref_standardattributes.asp) with special user agent behavior — the library can populate them for me using my property declarations. Otherwise I have to rewrite this functionality myself to populate them (yuck). IMO we don't want to be in the business of determining which attribute names to disallow; the developer should make this determination

klebba:
Also see #77

codeStryke:
The problem is the library makes the assumption that any custom property defined by the component author needs to be available as an attribute as well. Instead the author should be free to decide what the interface is, explicitly.

reliable-casey:
Can you explain the use case or issue you're working around?

codeStryke:
I have a lot of properties in the interface with names like `dataQuery`, `dataQueryVersion`or `dataQuery<xyz>`. Then the console is flooded with messages like these
![image](https://user-images.githubusercontent.com/5248016/108612373-243d7800-739d-11eb-87d9-5589eedeb283.png)

klebba:
@codeStryke what if the warning was just removed?

codeStryke:
@klebba the warnings may be potentially useful to resolve subtle hard to debug issues (not encountered any so far in 2yrs of using the framework). 
Isn't the root of the issue the automatic kebab case conversion (of property names) + automatic observing off attribute changes?

klebba:
[...]

PR #82, opened 2021-02-23
Removes all “nudges” from “x-element”.

### LINKED ISSUE #77: "id" behavior not guaranteed
`Unexpected key "FooElement.properties.id" shadows inherited interface, behavior not guaranteed.`

This warning is somewhat frustrating when I've purposely shadowed `id` in order to observe and react to changes on this property. I would choose another property name, but deep linking to an anchor like `page.htlm#id` is intrinsic behavior that I can't easily recreate. The easy workaround is to mirror the attribute for my own purposes, e.g. `<element id="foo" my-id="foo" ...>`

theengineear:
There’s been a lot of discussion about the opinions “x-element” makes to
be helpful — and sometimes the accusation is that “x-element” is getting
in the way.

This PR removes _all_ such nudges. This should feel inline with our
choice to remove errors for attempting to access `internal` properties
on the host (e.g., `host.myInternalProp` just returns `undefined` now).

In particular, we no longer error for the following:

* If an attribute is set for an unserializable property — we no longer
  observe such attributes at all.

And, we no longer warn for the following:

* If you have a property that shadows the HTMLElement interface’s
  prototype chain — e.g., “title”, “draggable”, “role”, “dataset”,
  “style”, etc, etc. The assumption is that users of “x-element” are
  experts and they would only do this _on purpose_.
* If you have a property that shadows “data-*” attributes.
* If you have a property that shadows “aria-*” attributes.

Closes #77.

### DIFF
--- a/x-element.js
+++ b/x-element.js
@@ -292,7 +292,7 @@ export default class XElement extends HTMLElement {
     const attributes = new Set();
     const inputMap = new Map();
     for (const [key, property] of entries) {
-      if (!property.internal) {
+      if (!property.internal && XElement.__serializableTypes.has(property.type)) {
         // Attribute names are case-insensitive — lowercase to properly check for duplicates.
         const attribute = property.attribute ?? XElement.__camelToKebab(key);
         XElement.__validatePropertyAttribute(constructor, key, property, attribute);
@@ -344,15 +344,8 @@ export default class XElement extends HTMLElement {
         throw new Error(`Unexpected value for "${path}.${subKey}" (expected Boolean, got ${typeName}).`);
       }
     }
-    if (!internal && RESERVED_PROPERTY_NAMES.has(key)) {
-      const { xElementPropertyName, reservedAttribute } = RESERVED_PROPERTY_NAMES.get(key);
-      if (xElementPropertyName) {
-        throw new Error(`Unexpected key "${path}" shadows in XElement.prototype interface.`);
-      } else if (reservedAttribute) {
-        console.warn(`Unexpected key "${path}" shadows related reserved attribute "${reservedAttribute}", behavior not guaranteed.`); // eslint-disable-line no-console
-      } else {
-        console.warn(`Unexpected key "${path}" shadows reserved interface, behavior not guaranteed.`); // eslint-disable-line no-console
-      }
+    if (!internal && XElement.__prototypeInterface.has(key)) {
+      throw new Error(`Unexpected key "${path}" shadows in XElement.prototype interface.`);
     }
     if (Reflect.has(property, 'attribute') && XElement.__typeIsWrong(String, attribute)) {
       const typeName = XElement.__getTypeName(attribute);
@@ -386,8 +379,12 @@ export default class XElement extends HTMLElement {
         }
       }
     }
-    if (reflect && (Reflect.has(property, 'type') === false || XElement.__serializableTypes.has(property.type) === false)) {
-      const typeName = property.type?.prototype && property.type?.name ? property.type.name : XElement.__getTypeName(property.type);
+    const unserializable = XElement.__serializableTypes.has(property.type) === false;
+    const typeName = property.type?.prototype && property.type?.name ? property.type.name : XElement.__getTypeName(property.type);
+    if (attribute && unserializable) {
+      throw new Error(`Found unserializable "${path}.type" (${typeName}) but "${path}.attribute" is defined.`);
+    }
+    if (reflect && unserializable) {
       throw new Error(`Found unserializable "${path}.type" (${typeName}) but "${path}.reflect" is true.`);
     }
     if (compute && !input) {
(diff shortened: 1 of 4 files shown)


theengineear:
@klebba , @charricknflx , @codeStryke , @erahhal — here's a controversial PR to bike shed on. We added these errors/warnings to try and be helpful. However, it seems like they might actually just be getting in the way. It's no sweat to me if we don't have them any more, but I wanted to know what ya'll think.

One thing we might do is add some lines of documentation that explains why you should familiarize with the "global" attributes, "data-*" attributes, "aria-*" attributes, and global properties as defined in the W3C specifications.

For example, you're not longer going to get warnings about interesting edge cases like the `class` attribute and it's associated `className` — however, my guess is that all of us are already intimately familiar and it's not the goal of "x-element" to teach us such nuance.

codeStryke:
I would still like to argue that we change the framework defaults to any one of the following
- not automatically observe for changes on attributes for properties (allow the author to use the `attribute` key in property definition) OR
- not change the property name to kebab case for the observed attribute name and instead just lower case the property name

AND leave the nudges in place.

klebba:
@theengineear what is the path forward here? This one fell off my radar but I would be glad to help push it over the finish line. It also sounds like @codeStryke has objections to the removal of nudges & has other requested changes (probably need to open a separate issue to discuss)

theengineear:
Yah! This issue did seem to fall off. My current opinion is **to remove all the nudges** as proposed here. And to be clear _all the nudges_ we're talking about are really things that are outside the purview of XElement — e.g., "data-*" attributes, etc.

The goal of XElement is not to warn you if you're misusing the platform. We only need to warn you if you're misusing the XElement interface.

As for the attributes/properties behavior discussion. There's already an issue for that — checkout #80. Let's have that conversation there and only discuss nudges here.

codeStryke:
It's fine you can remove the nudges, no more opposition from me

theengineear:
@klebba — I think we're all good to merge on this one. If anything, it might be good to take one more look at the top-level description to make sure you're OK with the change. You're welcome to take a pass at the code itself, mostly we just remove a bunch of construction-time and run-time checks 👌

klebba:
@theengineear Thanks for keeping me in the loop! Your succinct phrasing here really captures the intent of the change:

> The goal of XElement is not to warn you if you're misusing the platform. We only need to warn you if you're misusing the XElement interface.

Can you clarify this statement, I'm not sure I follow:

> If an attribute is set for an unserializable property — we no longer observe such attributes at all.

I gave the change set a once over; all in favor of removing run-time checks and slimming down element boot is great too. Great work!

x-element.js
```diff
@@ -292,7 +292,7 @@ export default class XElement extends HTMLElement {
     const attributes = new Set();
     const inputMap = new Map();
     for (const [key, property] of entries) {
-      if (!property.internal) {
+      if (!property.internal && XElement.__serializableTypes.has(property.type)) {
```
klebba, line 295: @theengineear can you clarify what is changing here?

x-element.js
```diff
@@ -344,15 +344,8 @@ export default class XElement extends HTMLElement {
...
-    if (!internal && RESERVED_PROPERTY_NAMES.has(key)) {
-      const { xElementPropertyName, reservedAttribute } = RESERVED_PROPERTY_NAMES.get(key);
-      if (xElementPropertyName) {
-        throw new Error(`Unexpected key "${path}" shadows in XElement.prototype interface.`);
-      } else if (reservedAttribute) {
-        console.warn(`Unexpected key "${path}" shadows related reserved attribute "${reservedAttribute}", behavior not guaranteed.`); // eslint-disable-line no-console
-      } else {
-        console.warn(`Unexpected key "${path}" shadows reserved interface, behavior not guaranteed.`); // eslint-disable-line no-console
-      }
+    if (!internal && XElement.__prototypeInterface.has(key)) {
```
klebba, line 347: If I understand correctly, this guards against accidentally stomping XElement internal names?

theengineear:
>Can you clarify this statement, I'm not sure I follow:

>>If an attribute is set for an unserializable property — we no longer observe such attributes at all.

Previously, we would observe attributes for properties which were not serializable. We did this solely for the purpose to erroring if you tried to interact with the attribute. Now, instead of erroring, we just do nothing.

```
// Before.
element.setAttribute('non-serializable-property'); // Error: Unexpected deserialization...

// After.
element.setAttribute('non-serializable-property'); // No more error, nothing really happens at all.
```

x-element.js
klebba: @theengineear thanks for the clarification in the main thread — I understand the conceptual change but not the line level change. It seems that we're introducing a more explicit check that wasn't previously in place, that's all I'm wondering about.

klebba:
@theengineear changes look great, no apprehension about merging here, was only trying to understand code-level changes to satisfy my own curiosity. Nice work!

x-element.js
theengineear: Ah, sorry I misunderstood! This part of attribute validation only applies to properties with attributes. Previously, only internal properties did not have an associated attribute. NOW, because non-serializable properties no longer have attributes, there's no need to check them either.

PR #85, opened 2021-05-07
Allow authors to extend “observedAttributes”.

### LINKED ISSUE #84: Allow extending `observedAttributes`.
Super minor change, but we expect all `observedAttributes` to come from our statically-declared `properties`. Instead, we should assume that authors may want to observe additional attributes beyond what they declared in the properties block. That is, we should allow this:

```
static get observedAttributes() {
  return [...super.observedAttributes, 'some-other-thing'];
}
```

theengineear:
Closes #84.

### DIFF
--- a/x-element.js
+++ b/x-element.js
@@ -127,7 +127,9 @@ export default class XElement extends HTMLElement {
   /** Extends HTMLElement.prototype.attributeChangedCallback. */
   attributeChangedCallback(attribute, oldValue, value) {
     const { attributeMap } = XElement.__constructors.get(this.constructor);
-    attributeMap.get(attribute).sync(this, value, oldValue);
+    // Authors may extend "observedAttributes". Optionally chain to account for
+    // attributes which we don't know about.
+    attributeMap.get(attribute)?.sync(this, value, oldValue);
   }
 
   /** Extends HTMLElement.prototype.adoptedCallback. */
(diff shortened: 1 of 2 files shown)


theengineear:
Small (one character) change to allow extension of observedAttributes. I found myself needing this functionality to track attribute changes without needing to define a related property. It essentially let's you go into "manual mode" with certain properties.

theengineear:
FYI @klebba — probably going to merge this in today. I'm gonna look into updating our dependencies today as well.

theengineear:
@klebba — Just bumping this one up for you. Any hesitations here or in the linked ticket?

x-element.js
```diff
@@ -126,7 +126,7 @@ export default class XElement extends HTMLElement {
   /** Extends HTMLElement.prototype.attributeChangedCallback. */
   attributeChangedCallback(attribute, oldValue, value) {
     const { attributeMap } = XElement.__constructors.get(this.constructor);
-    attributeMap.get(attribute).sync(this, value, oldValue);
+    attributeMap.get(attribute)?.sync(this, value, oldValue);
```
klebba, line 129: it would be good to leave a note here about why we're optional chaining here (else it looks less deliberate and more defensive)

klebba:
LG2M — thanks!

klebba (approved):

x-element.js
theengineear: Yep. Makes sense 👌

PR #90, opened 2021-07-30
Test “initial” works as expected with non-literal.

### LINKED ISSUE #89: Property definition "initial" value is unexpectedly reused across instances
For example, given an element `x-baz` with property block as follows:

```
static get properties() {
  return {
    bar: {
        type: Foo,
        initial: () => new Foo(),
    },
  };
}
```

Newly created instances of `<x-baz>` will reuse the same instance of `Foo`. In my case this creates an issue when attempting to destroy unwanted copies of `x-baz`. Because the reference is shared, any newly created `<x-baz>` will point to the original `Foo`

Happy to create a repro case if it helps

theengineear:
When “initial” is provided with a non-literal (compound) value, the
return value should never be shared across other instances of the
element.

Closes #89

### DIFF
--- a/test/test-initial-properties.js
+++ b/test/test-initial-properties.js
@@ -72,6 +72,29 @@ class TestElementStatic extends XElement {
 }
 customElements.define('test-element-static', TestElementStatic);
 
+class Compound {
+  get foo() {
+    return 'foo';
+  }
+}
+
+class TestElementCompound extends XElement {
+  static get properties() {
+    return {
+      compound: {
+        type: Compound,
+        initial: () => new Compound(),
+      },
+    };
+  }
+  static template(html) {
+    return ({ compound }) => {
+      return html`<div>${compound.foo}</div>`;
+    };
+  }
+}
+customElements.define('test-element-compound', TestElementCompound);
+
 it('basic initial properties', () => {
   const el = document.createElement('test-element-basic');
   document.body.append(el);
@@ -142,6 +165,15 @@ it('anonymous initial properties (predefined attributes)', () => {
   el.setAttribute('two', 'TWO');
   document.body.append(el);
   assert(el.shadowRoot.textContent === 'ONE, TWO');
+});
+
+it('compound initial properties are not shared accross element instances', () => {
+  const el1 = document.createElement('test-element-compound');
+  const el2 = document.createElement('test-element-compound');
+  document.body.append(el1, el2);
+  assert(el1.shadowRoot.textContent === 'foo');
+  assert(el2.shadowRoot.textContent === 'foo');
+  assert(el1.compound !== el2.compound);
 });
 
 it('cannot set initial to a bad type', () => {


theengineear:
@klebba — I'm having a hard time reproducing this issue in a test. Can you help be understand how this test isn't proving that x-element functions the way you want?

test/test-initial-properties.js
```diff
@@ -144,6 +167,15 @@ it('anonymous initial properties (predefined attributes)', () => {
   assert(el.shadowRoot.textContent === 'ONE, TWO');
 });
 
+it('compound initial properties are not shared accross element instances', () => {
+  const el1 = document.createElement('test-element-compound');
+  const el2 = document.createElement('test-element-compound');
+  document.body.append(el1, el2);
+  assert(el1.shadowRoot.textContent === 'foo');
+  assert(el2.shadowRoot.textContent === 'foo');
+  assert(el1.compound !== el2.compound);
```
theengineear, line 176: This is the behavior that you're looking for, right?

klebba: Yep — it looks like something is wrong on my end. I'll see if I can isolate a reproduction case

klebba:
@theengineear Thanks for checking this out; seems I need to take another look!

klebba:
Repro case:

```
<!doctype html>
<html>
    <head>
        <title>x-element initial</title>
        <style>:root { color-scheme: dark }</style>
    </head>
    <body>
        x-element issue <a href="https://github.com/Netflix/x-element/issues/89">#89</a>
        <div>
            <x-foo></x-foo>
            <x-foo></x-foo>
        </div>
        <script type="module">
            import XElement from 'https://cdn.jsdelivr.net/npm/@netflix/x-element@1.0.0-rc.43/x-element.js';

            class Baz {
                constructor() {
                    this.id = Math.floor(Math.random() * 100);
                }
            }

            export default class XFooElement extends XElement {
                static get properties() {
                    return {
                        bar: {
                            type: Baz,
                            internal: true,
                            initial: () => new Baz(),
                        },
                    };
                }

                static template(html) {
                    return ({ bar }) => html`
                        <div id="container">${ bar.id }</div>
                    `;
                }
            }

            customElements.define('x-foo', XFooElement);
        </script>
    </body>
</html>
```

theengineear:
Ah... well, it would help if we released `x-element`. This has been fixed, but we were sorta waiting for lit `2.0` to happen — it never did... I'll release today.

This explains why my test case (running against latest `master`) didn't pick up this behavior but your example (running against the latest release) does!

theengineear:
I'll release today so you can pull in the newest behavior.

theengineear:
@klebba — Gonna merge this in since I think it's a good test case to have locked down.

test/test-initial-properties.js
theengineear: As discussed, I think the confusion here is that we haven't released the changes that fix the issue you're seeing! I'm going to merge this and then release.

PR #100, opened 2021-11-20
Allow typeless props to be synced from attributes.

### LINKED ISSUE #99: Allow typeless properties to be set via an attribute.
The goal of typeless properties is to _just let me do it_. In this case, while it still doesn't make sense to `reflect` a typeless property, it's reasonable to allow an attribute to sync down to a typeless property.

I.e., consider the following element ...

```js
const MyElement extends XElement {
  static get properties() {
    return { data: {} };
  }
  static template(html) {
    return ({ data }) => html`<span>Here is some data: ${data}.</span>`;
  }
}
customElements.define('my-element', MyElement);
```

... I'd expect to be able to do the following ...

```html
<my-element data="123"></my-element>
```

Such that the output would read as:

```
Here is some data: 123.
```

However, because the property is typeless, "123" will never be synced from the _attribute_ to the _property_.

theengineear:
Closes #99.

### DIFF
--- a/x-element.js
+++ b/x-element.js
@@ -295,7 +295,7 @@ export default class XElement extends HTMLElement {
     const attributes = new Set();
     const inputMap = new Map();
     for (const [key, property] of entries) {
-      if (!property.internal && XElement.__serializableTypes.has(property.type)) {
+      if (XElement.__propertyHasAttribute(property)) {
         // Attribute names are case-insensitive — lowercase to properly check for duplicates.
         const attribute = property.attribute ?? XElement.__camelToKebab(key);
         XElement.__validatePropertyAttribute(constructor, key, property, attribute);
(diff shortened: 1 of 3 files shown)


theengineear:
Neat. The test just failed due to missing coverage.

Screen Shot 2021-11-19 at 5 08 17 PM

x-element.js
```diff
@@ -447,7 +447,7 @@ export default class XElement extends HTMLElement {
   // Called once per-property during constructor analysis.
   static __mutateProperty(constructor, propertyMap, key, property) {
     property.key = key;
-    property.attribute = !property.internal && XElement.__serializableTypes.has(property.type)
+    property.attribute = !property.internal && (XElement.__serializableTypes.has(property.type) || !property.type)
```
klebba, line 450: This assignment ternary is becoming difficult to follow, can we unpack it a bit?

theengineear: Sure, I'll take a look at it 👌

PR #103, opened 2022-03-03
Publishes tests and demos

### LINKED ISSUE #101: Include /test and /demo in the published version of x-element
Including the test suite is necessary for integrated testing. Might as well include the demos too

klebba:
Closes #101

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,7 +19,8 @@
   },
   "files": [
     "/x-element.js",
-    "/x-element-bundle.js",
+    "/demo",
+    "/test",
     "/etc"
   ],
   "dependencies": {


klebba:
@theengineear mind taking a look? Should be a pretty simple change but will require publication of a new version

klebba:
As an integrator I actually do want to be able to run all the tests in my application environment. In certain development contexts I need to demonstrate that I can run vendor tests as well.

In my experience integrators skirt running 3rd party tests because they are too difficult to run without a major investment (require extra dependencies, special environments, compilation, etc). However in our case we have taken special care to ensure that `x-element` and `x-test` test are dead simple to run and work in the browser without any of these steps. So I would ask instead: is there a good reason not to include them in the distro? Thanks for considering

theengineear:
>Is there a good reason not to include them in the distro?

Sure, I think this is the key question. I think my arguments would be as follows:

* As the contents of `/test` and `/demo` (which are not required to use the element) gets _large_, download speeds over network will also increase†
* As the contents of `/test` and `/demo` (which are not required to use the element) gets _large_, space on disk will also increase††

† To which you might argue, (1) the size of `/test` and `/demo` is unlikely to get _that large_ and (2) this is what _caching_ is for.

†† To which you might argue, (1) the size of `/test` and `/demo` is unlikely to get _that large_ and (2) if you're worried about disk space, you could just manually delete these things before committing them.

So, it would seem that _not adding these files_ is getting in your way without any _real_ benefit — I'm convinced.

theengineear (approved):
Thanks for considering my reservations here — I think the value you proposed ultimately outweighs my concerns. I.e., it's easier as an integrator to hide/delete these files if they're present vs. magically try and retrieve these files if they're omitted.

klebba:
Awesome, thanks for working through that. I'm not able to merge so please do the deed when you're ready

package.json
```diff
@@ -19,7 +19,8 @@
   },
   "files": [
     "/x-element.js",
-    "/x-element-bundle.js",
```
klebba, line 22: 👋🏻

PR #111, opened 2022-04-12
Add Custom Elements Everywhere test suite.

### LINKED ISSUE #106: Ensure x-element compatibility with Custom Elements Everywhere test suite
https://custom-elements-everywhere.com

theengineear:
In the future, we may choose to add this test suite to the
custom-elements-everywhere repo. However, we should make sure that our
current test suite covers this need so that we don’t somehow drift as
we make changes.

Closes #106

---

Here's a visual on the test suite for reference (grabbed from the custom elements everywhere page).

Screen Shot 2022-04-11 at 4 31 27 PM

### DIFF
--- a/test/index.js
+++ b/test/index.js
@@ -22,3 +22,4 @@ test('./test-computed-properties.html');
 test('./test-observed-properties.html');
 test('./test-listeners.html');
 test('./test-scratch.html');
+test('./test-custom-elements-everywhere.html');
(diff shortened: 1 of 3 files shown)


klebba:
Wow, neato — yes I think this looks like the best approach here — I would guess that we score 100% right?

theengineear:
>I would guess that we score 100% right?

Yah, for sure.

theengineear:
Moved to #118.

PR #112, opened 2022-04-12
Leverage class fields and private class fields.

### LINKED ISSUE #63: Make internals of x-element #private when support lands in modern browsers.
In #58, we'll be refactoring our code to have a clearer distinction between what is "public facing" and what is "private facing". We've put double underscores in front of methods which we intend to truly internalize later.

Ultimately, this ticket is just a reminder to switch the `__` in our code to `#` when it's supported in browsers we target.

theengineear:
The “Class Fields” proposal has been adopted as part of ECMAScript 2022
and tests pass leveraging this feature in Chrome, Firefox, and Safari.

Closes #63.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -31,7 +31,7 @@ module.exports = {
     browser: true,
   },
   parserOptions: {
-    ecmaVersion: 2020,
+    ecmaVersion: 2022,
     sourceType: 'module',
   },
 };
--- a/package.json
+++ b/package.json
@@ -29,7 +29,7 @@
   "devDependencies": {
     "@netflix/element-server": "^1.0.12",
     "@netflix/x-test": "^1.0.0-rc.16",
-    "eslint": "^7.5.0",
+    "eslint": "^8.13.0",
     "puppeteer": "^13.4.0",
     "tap-parser": "^10.0.1"
   }
--- a/x-element.js
+++ b/x-element.js
@@ -14,14 +14,12 @@ import { unsafeHTML } from '../../lit-html/directives/unsafe-html.js';
 import { unsafeSVG } from '../../lit-html/directives/unsafe-svg.js';
 import { until } from '../../lit-html/directives/until.js';
 
-// TODO: #63: replace "__ >> #" when private fields are supported.
-
 /** Base element class for creating custom elements. */
 export default class XElement extends HTMLElement {
   /** Extends HTMLElement.observedAttributes to handle the properties block. */
   static get observedAttributes() {
-    XElement.__analyzeConstructor(this);
-    return [...XElement.__constructors.get(this).attributeMap.keys()];
+    XElement.#analyzeConstructor(this);
+    return [...XElement.#constructors.get(this).attributeMap.keys()];
   }
 
   /**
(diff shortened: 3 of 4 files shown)


theengineear:
@klebba — hesitations or reservations? I ran our test suite in Chrome, Firefox, and Safari after this change. Note, this change is essentially just replacing `__` with `#` (for context, we've been using a `__` for the express purpose of making this particular PR simple to implement when the time came).

.eslintrc.js
```diff
@@ -31,7 +31,7 @@ module.exports = {
     browser: true,
   },
   parserOptions: {
-    ecmaVersion: 2020,
+    ecmaVersion: 'latest',
```
theengineear, line 34: I'm not sure that we need to pin `2020` anymore — I only ended up momentarily scratching my head when bumping the eslint version failed to fix my linting issues. If you think it'd be better, we can just re-pin to `2022` rather than `'latest'` 🤷

package.json
```diff
@@ -29,7 +29,7 @@
   "devDependencies": {
     "@netflix/element-server": "^1.0.12",
     "@netflix/x-test": "^1.0.0-rc.16",
-    "eslint": "^7.5.0",
+    "eslint": "^8.13.0",
```
theengineear, line 32: The class fields support only landed as part of eslint 8+.

x-element.js
```diff
@@ -894,39 +892,38 @@ export default class XElement extends HTMLElement {
...
-}
 
-// TODO: #63: define as private class fields inside the constructor when supported.
-XElement.__constructors = new WeakMap();
-XElement.__hosts = new WeakMap();
-XElement.__propertyKeys = new Set(['type', 'attribute', 'input', 'compute', 'observe', 'reflect', 'internal', 'readOnly', 'initial', 'default']);
-XElement.__serializableTypes = new Set([Boolean, String, Number]);
-XElement.__caseMap = new Map();
-XElement.__prototypeInterface = new Set(Object.getOwnPropertyNames(XElement.prototype));
+  static #constructors = new WeakMap();
```
theengineear, line 923: Note that these are now defined _within_ the class, rather than outside-and-after. This was always the goal.

klebba:
@theengineear what would this do to our min supported browser versions? I'm totally supportive of adopting the new syntax, but would not want to ice out any recent browser incarnations unnecessarily

.eslintrc.js
klebba: I suppose I would prefer to use an explicit version so that it's pinned at some moment in time

x-element.js
klebba: cooooool

theengineear:
>What would this do to our min supported browser versions?

Here's a screen grab of the relevant support from caniuse:

Screen Shot 2022-04-12 at 5 16 29 PM

Screen Shot 2022-04-12 at 5 16 45 PM

.eslintrc.js
theengineear: Sure, I changed it to `2022`.

theengineear:
Ah, nice — that breakdown is super helpful. Thanks! This doesn't bother me terribly to move forward here. For our uses-cases, it's a non-issue. It's not a huge deal to wait if you want though.

If we _do_ want to wait, we should probably come to a consensus on what year we attempt to target? I.e., "we target the ECMAScript version `currentYear - 2`?

theengineear:
☝️ — @klebba, still want to pause on this change? And if so, did you have a sense of the ECMAScript year you were hoping to target?

klebba:
@theengineear Thanks for picking this up again — I think we should go for it!

theengineear:
Great! Gonna merge and update other branches.

PR #116, opened 2022-04-16
Template engine

theengineear:
This is a first pass at what it might look like to ship a templating engine as part of `x-element` such that we can hit our goal of _zero dependencies_.

### DIFF
--- a/demo/lit-html/base-element.js
+++ b/demo/lit-html/base-element.js
@@ -1,0 +1,28 @@
+import XElement from '../../x-element.js';
+import { asyncAppend } from 'https://unpkg.com/lit-html/directives/async-append.js?module';
+import { asyncReplace } from 'https://unpkg.com/lit-html/directives/async-replace.js?module';
+import { cache } from 'https://unpkg.com/lit-html/directives/cache.js?module';
+import { classMap } from 'https://unpkg.com/lit-html/directives/class-map.js?module';
+import { directive } from 'https://unpkg.com/lit-html/directive.js?module';
+import { guard } from 'https://unpkg.com/lit-html/directives/guard.js?module';
+import { html, render as originalRender, svg } from 'https://unpkg.com/lit-html/lit-html.js?module';
+import { ifDefined } from 'https://unpkg.com/lit-html/directives/if-defined.js?module';
+import { live } from 'https://unpkg.com/lit-html/directives/live.js?module';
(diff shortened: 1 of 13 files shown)


test/test-template-engine.js
```diff
@@ -0,0 +1,671 @@
+import XElement from '../x-element.js';
```
theengineear, line 1: New set of tests to spec out the functionality we want from our default template engine.

package.json
```diff
@@ -23,13 +23,11 @@
     "/test",
     "/etc"
   ],
-  "dependencies": {
-    "lit-html": "^2.2.0"
-  },
+  "dependencies": {},
```
theengineear, line 26: Yay for zero dependencies. This is a key goal of this change set.

demo/lit-html/base-element.js
```diff
@@ -0,0 +1,28 @@
+import XElement from '../../x-element.js';
```
theengineear, line 1: Example of how to get back previous behavior of binding `lit-html` if desired. This is a key test-case to make sure that migration won't require changing _every component at once_ within a codebase.

demo/uhtml/base-element.js
```diff
@@ -0,0 +1,9 @@
+import XElement from '../../x-element.js';
```
theengineear, line 1: Example of how we can now bind other template engines like `µhtml`.

test/test-initialization-errors.js
```diff
@@ -92,10 +92,10 @@ it('errors are thrown in connectedCallback when template result fails to render'
         strings: {},
       };
     }
-    static template(html, { repeat }) {
+    static template(html, { keyedMap }) {
```
theengineear, line 95: The required changes to the current test suite were _super minimal_. Really, it's only due to slight differences in how keyed lists are managed.

This doesn't mean that this isn't a large breaking change — if anything it just reflects the fact that the behavior of x-element should really be decoupled from the underlying templating engine implementation.

theengineear:
OK @klebba — This is by no means a finished product, but I was curious to try and get a POC of what rolling our own templating engine might actually look like.

Have a look. Or, don't — I'm likely going to let this sit on the back burner for a while now 🤷 

The TL;DR is that it was _much easier than anticipated_ code the bulk of the behavior up.

theengineear:
Moved to #119.

PR #119, opened 2022-09-02
Decouple templating engine from “x-element”.

### LINKED ISSUE #105: Link to hosted x-element build in README
I'm not sure what the best options are lately, but something like this could help fast adoption & fiddling:

https://unpkg.com/browse/@netflix/x-element@1.0.0-rc.49/x-element.js

### LINKED ISSUE #113: Publish a playground for simple fiddling with “x-element”.
We should host static page (perhaps within the GitHub ecosystem) to allow fiddling with “x-element”. It'd be interesting to see how light-weight a solution we could come up with which would still essentially give folks a light-weight text editor and rendered output.

Not sure on exactly the where / how yet, but I think this would be valuable to lower the barrier-to-entry.

### LINKED ISSUE #115: Decouple templating engine and ship with a minimal default.
Decoupling rendering logic from core `x-element` logic and rolling our own internal templating engine would provide the following benefits:

1. **Zero Dependencies** — This makes installation, integration, and consumption of this library incredibly straightforward and tightly controlled.
2. **Simple by Default** — A major goal of this project is to keep the interface size to a minimum. It's not meant to _do it all_. By rolling a custom templating engine, we can (1) expose only functionality that we think is necessary and (2) allow integrators to pick their own templating engine if ours doesn't meet their needs (e.g., [lit-html](https://github.com/lit/lit/tree/main/packages/lit-html) or [µhtml](https://github.com/WebReflection/uhtml)).

**How might the interface change?**

This list is incomplete and subject to change — but, to give an idea:

* *Fewer built-in template utilities.* Over many years of developing Web Components, we've grown to favor a pattern where attribute/property/content values are managed by the element such that the _render loop is incredibly simple and lightweight_.
* *Less syntactic sugar.* There are really only three things a templating engine needs to consider — attributes, properties, and content. As long as there is a way to inject these core things, everything else can be managed outside the templating engine. This may prevent blurring the lines between separate areas of concern. For example, the `?` and `@` behavior may not strictly be necessary.
* *Stricter interpolation patterns.* By more clearly defining what is/isn't valid interpolation, we can nudge integrators into cleaner code. For example `attr="foo-${value}-bar"` is quite complicated to read and difficult to internally manage — instead, we could insist it be written as `attr="${value}"`, where value is computed elsewhere to be `foo-${userInput}-bar` or some such. Additionally, we could consider not supporting interpolation within `<style>` tags _at all_.

### LINKED ISSUE #121: Discussion: Add syntactic sugar for boolean attribute bindings to Template Engine.
Authors can already leverage a `boolean` updater to accomplish a boolean attribute binding. Adding syntactic sugar may potentially be a more ergonomic solution.

**Why we might add it:**

* It may be expected behavior at this point (as it exists in other templating engines)
* We have leveraged this syntactic sugar in the past, so there’s some precedent for us.

**Why we might not add it:**

* It’s redundant — you can already achieve this behavior.
* It adds complexity / size to the base element.
* It adds an additional runtime check — this may or may not be a performance drawback.
* Conceptually, it blurs the lines between “what to do” and “how to do it”. We may not want to cross that line.

### Proposal

Users would be able to declare something like the following to bind an attribute as a boolean (see the `?` in there?):

```js
static template(html) {
  return ({ trueOrFalse }) => {
    return html`<div ?data-true="${trueOrFalse}">Hi There.</div>`;
  };
}
```

### Currently

…for comparison, this is how you would do this currently.

```js
static template(html, { boolean }) {
  return ({ trueOrFalse }) => {
    return html`<div data-true="${boolean(trueOrFalse)}">Hi There.</div>`;
  };
}
```

Note that the _what to do_ (i.e., bind an attribute) is nicely separated from the _how to do it_ (i.e., in a boolean fashion).

### LINKED ISSUE #122: Add performance testing for template engine.
We should find a way to set up performance testing for our template engine to prove that it matches up to similar templating engines.

### LINKED ISSUE #124: Link to SPEC.md from README.md
Title says it all.

### LINKED ISSUE #131: Add TypeScript declaration file (e.g., “index.d.ts”) to declare interface.
Title says it all. Doing this will (1) help formalize our interface and (2) make us more compatible.

theengineear:
This makes the following changes:                                        
                                                                         
1. The library no longer declares _any_ dependencies.                    
2. A new, default template engine is provided.                           
3. Integrators may self-configure the templating engine to be used.      
4. Adds TypeScript “x-element.d.ts” declarations.                        
                                                                         
Closes #105. Closes #113. Closes #115. Closes #121. Closes #122.         
Closes #124. Closes #131.    

---

To get back to _exactly_ the prior interface, you can import `lit-html` and plug it back in yourself:

```javascript
import XElement from '../../x-element.js';
import { asyncAppend } from '../../lit-html/directives/async-append.js';
import { asyncReplace } from '../../lit-html/directives/async-replace.js';
import { cache } from '../../lit-html/directives/cache.js';
import { classMap } from '../../lit-html/directives/class-map.js';
import { directive } from '../../lit-html/directive.js';
import { guard } from '../../lit-html/directives/guard.js';
import { html, render as originalRender, svg } from '../../lit-html/lit-html.js';
import { ifDefined } from '../../lit-html/directives/if-defined.js';
import { live } from '../../lit-html/directives/live.js';
import { repeat } from '../../lit-html/directives/repeat.js';
import { styleMap } from '../../lit-html/directives/style-map.js';
import { templateContent } from '../../lit-html/directives/template-content.js';
import { unsafeHTML } from '../../lit-html/directives/unsafe-html.js';
import { unsafeSVG } from '../../lit-html/directives/unsafe-svg.js';
import { until } from '../../lit-html/directives/until.js';

export default class BaseElement extends XElement {
  // Use lit-html's template engine rather than the built-in x-element engine.
  static get templateEngine() {
    const render = (container, template) => originalRender(template, container);
    return {
      render, html, svg, asyncAppend, asyncReplace, cache, classMap, directive,
      guard, ifDefined, live, repeat, styleMap, templateContent, unsafeHTML,
      unsafeSVG, until,
    };
  }
}
```


theengineear:
>Bail out! Coverage goal of 100% for http://0.0.0.0:8080/node_modules/@netflix/x-element/x-element.js not met (got 99.90%).

I forgot that we have a requirement of 100% coverage 🎉 

Fixing now.

klebba (dismissed):
Excellent work! Looking forward to testing this out. Feel free to merge any time

theengineear:
Alright, after all that suspense — going to merge (based on prior approval). Zero dependencies, here we come!

PR #126, opened 2023-01-29
Setup listeners before initial render.

theengineear:
Previously, “initializeHost” was called ahead of “addListeners”. There was an edge case where such depth-first initialization made it possible for synchronous events from children to be emitted before the delegated event listeners had been setup on the parent.

This change ensures the following happens in order on first connection:
1. The host itself is initialized.
2. The static listeners are added.
3. The host undergoes its initial render.

Previously, (2) and (3) were conceptually swapped.

### DIFF
--- a/test/test-listeners.js
+++ b/test/test-listeners.js
@@ -1,10 +1,23 @@
 import XElement from '../x-element.js';
 import { assert, it } from '../../../@netflix/x-test/x-test.js';
 
+class TestElementChild extends HTMLElement {
+  connectedCallback() {
+    const eventType = 'connected';
+    const eventData = { bubbles: true, composed: true };
+    this.dispatchEvent(new CustomEvent(eventType, eventData));
+  }
+}
+customElements.define('test-element-child', TestElementChild);
+
 class TestElement extends XElement {
   static get properties() {
     return {
       clicks: {
+        type: Number,
+        default: 0,
+      },
+      connections: {
         type: Number,
         default: 0,
       },
@@ -23,7 +36,7 @@ class TestElement extends XElement {
     };
   }
   static get listeners() {
-    return { click: this.onClick };
+    return { click: this.onClick, connected: this.onConnected };
   }
   static template(html) {
     return ({ clicks, count }) => {
@@ -32,6 +45,7 @@ class TestElement extends XElement {
         <button id="decrement" type="button">-</button>
         <span>clicks: ${clicks} count ${count}</span>
         <div id="custom-event-emitter"></div>
+        <test-element-child id="connected-event-emitter"></test-element-child>
       `;
     };
   }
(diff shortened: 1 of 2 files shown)


test/test-listeners.js
```diff
@@ -119,6 +136,19 @@ it('test manual lifecycle', () => {
   assert(count === 2, 'listener was removed');
 });
 
+it('test synchronous event handling', () => {
```
theengineear, line 139: Yes indeed, this test does fail without the related change in `x-element.js`. Writing this test made it additionally-clear that this is indeed something worth fixing, by the way.

x-element.js
```diff
@@ -705,8 +716,9 @@ export default class XElement extends HTMLElement {
         invalidProperties.add(property);
       }
       hostInfo.initialized = true;
-      XElement.#updateHost(host);
```
theengineear, line 708: It turned out that more nuanced handling was conceptually needed here. I had initially done the broad-brush-stroke thing and just swapped the lines in `connectedCallback` — however, I think it makes more sense to just _delay_ this `#updateHost` call until after we've set up our listeners.

You can see that there are some minor gymnastics involved now, but I think this is the most-correct thing we can do here.

test/test-listeners.js
```diff
@@ -1,13 +1,26 @@
 import XElement from '../x-element.js';
 import { assert, it } from '../../../@netflix/x-test/x-test.js';
 
+class TestElementChild extends HTMLElement {
+  connectedCallback() {
```
theengineear, line 5: Here's the test setup (note that this extends a basic `HTMLElement` by the way) — we have an element that emits an event as soon as it's connected to the DOM. Maybe not the _best_ pattern, but honestly, not completely pathological.

Seems fairly clear that we should try and support this.

x-element.js
```diff
@@ -118,8 +118,7 @@ export default class XElement extends HTMLElement {
 
   /** Extends HTMLElement.prototype.connectedCallback. */
   connectedCallback() {
-    XElement.#initializeHost(this);
-    XElement.#addListeners(this);
+    XElement.#connectHost(this);
```
theengineear, line 121: I wrapped these simply for readability / symmetry. All of these are private, so it's not like we're making public interface changes here.

x-element.js
```diff
@@ -686,6 +685,18 @@ export default class XElement extends HTMLElement {
   }
 
   // Called once per-host from initial "connectedCallback".
+  static #connectHost(host) {
+    const initialized = XElement.#initializeHost(host);
+    XElement.#addListeners(host);
```
theengineear, line 690: Again, this is the magic-foo here. See how we split apart _initialization_ and _initial render_? This allows us to sneak in _adding event listeners_ between these two points in time.

theengineear:
@klebba — Want to give this a review? It's slightly more nuanced than I had previously thought. But, I think this is the _correct_ factoring.

x-element.js
klebba: Makes sense to me — thanks for all the helpful context, made review a breeze!

klebba (approved):
LG2M — great work!

PR #130, opened 2023-03-01
Skip initial lookup for internal properties.

### LINKED ISSUE #129: Internal props should not participate in pre-upgrade element initialization.
The point of the `internal` properties are to specifically _not participate_ in the public element interface. I.e., `element.title` should have nothing to do with `element.internal.title`. This is usually benign, but can cause problems if the related public property has restrictions on how it’s meant to be used. For example, the following will fail if you register it and try to print it in the DOM:

```javascript
class InternalExample extends XElement {
  static get properties() {
    return { children: { internal: true } }; // The "children" property is _read only_ on the public interface.
  }
}
// TypeError: Cannot set property children of #<Element> which has only a getter
```

**Proposal**

Simply don’t perform a lookup on the public interface for internal properties. This is the intended behavior, but there’s just no guard yet.

theengineear:
Internal properties should be completely decoupled from the public interface. This change set adds a guard to ensure that we don’t try to lookup the property name on the public interface when upgrading the element.

Closes #129.

### DIFF
--- a/test/test-internal-properties.js
+++ b/test/test-internal-properties.js
@@ -15,6 +15,13 @@ class TestElement extends XElement {
         input: ['internalProperty'],
         compute: internalProperty => internalProperty,
       },
+      // This exists on the public interface, we want to ensure that there's no
+      //  issue using this same name on the internal interface.
+      children: {
+        // Use a type that's in conflict with the public interface.
+        type: Boolean,
+        internal: true,
+      },
     };
   }
   static template(html) {
@@ -42,7 +49,10 @@ it('can use "ownKeys" api.', () => {
   document.body.append(el);
   const ownKeys = Reflect.ownKeys(el.internal);
   assert(
-    ownKeys.length === 2 && ownKeys[0] === 'internalProperty' && ownKeys[1] === 'internalComputedProperty',
+    ownKeys.length === 3 &&
+    ownKeys[0] === 'internalProperty' &&
+    ownKeys[1] === 'internalComputedProperty' &&
+    ownKeys[2] === 'children',
     'The "ownKeys" trap does not work.'
   );
 });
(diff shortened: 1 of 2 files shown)


test/test-internal-properties.js
```diff
@@ -42,7 +49,10 @@ it('can use "ownKeys" api.', () => {
   document.body.append(el);
   const ownKeys = Reflect.ownKeys(el.internal);
   assert(
-    ownKeys.length === 2 && ownKeys[0] === 'internalProperty' && ownKeys[1] === 'internalComputedProperty',
+    ownKeys.length === 3 &&
+    ownKeys[0] === 'internalProperty' &&
+    ownKeys[1] === 'internalComputedProperty' &&
+    ownKeys[2] === 'children',
```
theengineear, line 55: This is sorta unrelated — because we added a new property on the test element, we need to also update this test to consider it.

x-element.js
```diff
@@ -735,16 +735,19 @@ export default class XElement extends HTMLElement {
     // Process possible sources of initial state, with this priority:
     // 1. imperative, e.g. `element.prop = 'value';`
     // 2. declarative, e.g. `<element prop="value"></element>`
-    const { key, attribute } = property;
+    const { key, attribute, internal } = property;
     let value;
     let found = false;
```
theengineear, line 740: Note that we will now _always_ return `found = false` when looking up internal properties on the pre-upgraded element’s public interface — because, obviously, they’re not on the public interface…

theengineear:
For context — I'd consider this a _bug_ with the current implementation. I don't think there should be any conceptual argument about the behavior here. We just need to add in this guard.

theengineear:
@klebba — This is a pretty small change and since we already caught up on it offline, I'm just going to merge it.

PR #135, opened 2023-03-16
Allow special characters in quoted attributes.

theengineear:
Characters like “<”, “/”, and “>” are special in html and we use them as boundary markers to determine where we are as we walk through the result of a tagged template literal callback.

However, it’s perfectly allowable to use the characters in the markup itself if they’re quoted. Take the following, resonable example:

```
html`<a href="https://example.com" target="${target}">click me</a>`
```

Previously, because of the “/” characters in the “href” attribute declaration, the “${target}” binding would get missed. This is now fixed to allow quoted, special characters.

### DIFF
--- a/x-element.js
+++ b/x-element.js
@@ -1062,9 +1062,9 @@ class Template {
   static #templates = new WeakMap();
   static #templateResults = new WeakMap();
   static #updaters = new WeakMap();
-  static #ATTRIBUTE = /<[a-zA-Z0-9-]+[^/<>]* ([a-z][a-z-]*)="$/;
-  static #BOOLEAN_ATTRIBUTE = /<[a-zA-Z0-9-]+[^/<>]* \?([a-z][a-z-]*)="$/;
-  static #PROPERTY = /<[a-zA-Z0-9-]+[^/<>]* \.([a-z][a-zA-Z0-9_]*)="$/;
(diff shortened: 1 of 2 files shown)


x-element.js
```diff
@@ -1062,9 +1062,9 @@ class Template {
   static #templates = new WeakMap();
   static #templateResults = new WeakMap();
   static #updaters = new WeakMap();
-  static #ATTRIBUTE = /<[a-zA-Z0-9-]+[^/<>]* ([a-z][a-z-]*)="$/;
-  static #BOOLEAN_ATTRIBUTE = /<[a-zA-Z0-9-]+[^/<>]* \?([a-z][a-z-]*)="$/;
-  static #PROPERTY = /<[a-zA-Z0-9-]+[^/<>]* \.([a-z][a-zA-Z0-9_]*)="$/;
+  static #ATTRIBUTE = /<[a-zA-Z0-9-]+(?:[^/<>]|"[^"]*")* ([a-z][a-z-]*)="$/;
+  static #BOOLEAN_ATTRIBUTE = /<[a-zA-Z0-9-]+(?:[^/<>]|"[^"]*")* \?([a-z][a-z-]*)="$/;
+  static #PROPERTY = /<[a-zA-Z0-9-]+(?:[^/<>]|"[^"]*")* \.([a-z][a-zA-Z0-9_]*)="$/;
```
theengineear, line 1067: The `"[^"]*"` says “allow bookended, quoted content.”

PR #136, opened 2023-03-17
Allow everything but “>” character in template.

theengineear:
Eventually, we will support this completely, but not at the cost of performance. The prior solution here encountered catastrophic backtracking due to how it was looking for escaped characters.

### DIFF
--- a/x-element.js
+++ b/x-element.js
@@ -1062,9 +1062,9 @@ class Template {
   static #templates = new WeakMap();
   static #templateResults = new WeakMap();
   static #updaters = new WeakMap();
-  static #ATTRIBUTE = /<[a-zA-Z0-9-]+(?:[^/<>]|"[^"]*")* ([a-z][a-z-]*)="$/;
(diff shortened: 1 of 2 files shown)


x-element.js
```diff
@@ -1062,9 +1062,9 @@ class Template {
   static #templates = new WeakMap();
   static #templateResults = new WeakMap();
   static #updaters = new WeakMap();
-  static #ATTRIBUTE = /<[a-zA-Z0-9-]+(?:[^/<>]|"[^"]*")* ([a-z][a-z-]*)="$/;
-  static #BOOLEAN_ATTRIBUTE = /<[a-zA-Z0-9-]+(?:[^/<>]|"[^"]*")* \?([a-z][a-z-]*)="$/;
-  static #PROPERTY = /<[a-zA-Z0-9-]+(?:[^/<>]|"[^"]*")* \.([a-z][a-zA-Z0-9_]*)="$/;
```
theengineear, line 1067: The combinatorics here are fighting against us. Likely, we'll want to do more to intelligently _walk_ over the strings in the template — just trying to keep this as simple as possible in our first pass with a templating engine.

PR #137, opened 2023-03-17
Update dependencies for testing.

theengineear:
Some notes:
* “x-test” comes with grouping features, so that’s leveraged a bit.
* “x-test” now produces TAP 14, so we can bump “tap-parser” as well.
* “puppeteer” bumps from 13.x >> 19.x — fortunately nothing broke us.


theengineear:
Just some internal changes (no interface changes) — going to get this one merged.

PR #138, opened 2023-03-19
Fix template injection for special characters.

theengineear:
The following should work (and work performantly) as a result of this change set:

```javascript
html`\
  <svg
    id="svg"
    class="<><></></>"
    xmlns="http://www.w3.org/2000/svg"
    viewBox="0 0 24 24"
    width="${ifDefined(width)}"
    height="${ifDefined(height)}">
    <circle id="circle" r="12" cx="12" cy="12"></circle>
  </svg>
```


theengineear:
As part of this work — I had to write a performance test to target this phase of templating. Not only does this change _fix_ the core issue (see the test that got fixed in the diff). It also more than _halved_ the time it takes to inject.

Finally, it now throws an error if things seem off during templating — we may want to fine-tune this in the future, but based on how we walk over the interpolated strings, we now sort of get the opportunity to error-out for free.

test/test-template-engine.js
```diff
@@ -379,10 +344,7 @@ describe('html rendering', () => {
     container.remove();
   });
 
-  // TODO: trying to find escaped values in strings is possible via a regex, but
-  //  making that performant is nuanced. I.e., it's easy to cause catastrophic
-  //  backtracking.
-  it.todo('renders elements with "<" or ">" characters in attributes', () => {
```
theengineear, line 385: This is the key thing that _now works_.

test/test-template-engine.js
```diff
@@ -761,6 +723,86 @@ describe('svg updaters', () => {
 });
 
 describe('rendering errors', () => {
+  describe('templating', () => {
+    it.skip('throws when attempting to interpolate within a style tag', () => {});
```
theengineear, line 727: Added a couple placeholders — we should make this even more strict. For a future PR though.

x-element.js
```diff
@@ -1207,6 +1223,34 @@ class Template {
     return reference;
   }
 
+  static #exhaustString(string, state) {
```
theengineear, line 1226: Rather than use a stateless regular expression to look back across the string — we now walk through each piece of the string with a state object. Although a bit more cumbersome, it's more performant and more correct.

theengineear:
This change set is limited to the default templating engine. I.e., not too much risk to merging here — going to get this one into `main`.

PR #139, opened 2023-03-21
Throw for sketchy interpolation.

theengineear:
If a user tries to interpolate in “style” or “script” tags, just bail out with a clear error. Previously, we sort of silently replaced the content, which is ultimately harder on developers to debug.

### DIFF
--- a/x-element.js
+++ b/x-element.js
@@ -1107,7 +1107,7 @@ class Template {
               html = html.slice(0, -3 - name.length) + `x-element-property-$${iii}="${name}`;
               state.index = 1; // Accounts for an expected quote character next.
             } else {
-              throw new Error(`Found invalid template near "${string.slice(state.index)}".`);
+              throw new Error(`Found invalid template string "${string}" at "${string.slice(state.index)}".`);
             }
           }
         } else {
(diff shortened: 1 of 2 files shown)


x-element.js
```diff
@@ -1107,7 +1107,7 @@ class Template {
               html = html.slice(0, -3 - name.length) + `x-element-property-$${iii}="${name}`;
               state.index = 1; // Accounts for an expected quote character next.
             } else {
-              throw new Error(`Found invalid template near "${string.slice(state.index)}".`);
+              throw new Error(`Found invalid template string "${string}" at "${string.slice(state.index)}".`);
```
theengineear, line 1110: Snuck this in as well — there are cases where having the additional, full string is valuable. I stopped short of including the entire template string as that can get rather long.

x-element.js
```diff
@@ -1278,7 +1278,7 @@ class Template {
             node.textContent = '';
             items.push({ path, key: contentMatch[1], type: 'text-content' });
           } else {
-            node.textContent = '/* x-element: Interpolation is not allowed here. */';
+            throw new Error(`Interpolation of "${node.localName}" tags is not allowed.`);
```
theengineear, line 1281: Just bail. It's much better to just error loudly here.

PR #144, opened 2023-05-24
Ensure computation on initialization.

### LINKED ISSUE #143: Fix edge case where computed property is used between construction and initialization.
Through manual manipulation via `document.createElement` — it’s possible to get into a bizarre scenario where computed properties appear to be inexplicably wrong. Take the following:

```
import XElement from 'https://deno.land/x/element/x-element.js';

class TestElement extends XElement {
  static get properties() {
    return {
      a: { type: Number },
      b: { type: Number },
      c: { type: Number, internal: true, input: ['a', 'b'], compute: (a, b) => a + b },
    };
  }
}
customElements.define('test-element', TestElement);

// Hint: This element is constructed, but not initialized.
const element = document.createElement('test-element');
element.setAttribute('a', '1');
element.setAttribute('b', '2');
console.log({ a: element.a, b: element.b, c: element.internal.c });
/* 
  This is reasonable — it's not expected that attributes sync before DOM connection.
  {
    a: undefined,
    b: undefined,
    c: NaN
  }
*/

// By attaching it to the DOM, we finally initialize.
document.body.append(element);
console.log({ a: element.a, b: element.b, c: element.internal.c });
/*
  This is NOT reasonable — everything should _just work_ after initialization.
  {
    a: 1,
    b: 2,
    c: NaN
  }
*/
```

To summarize — this edge case is caused by manual manipulation of attributes and introspection of downstream, internal, computed properties thereof.

In theory, this _might_ even be caused by race conditions due to timing issues between attribute setting, property introspection, and forcible upgrade paths via `customElements.upgrade`. However, the simple target is to fix the easily-reproducible issue.

**Proposal**

Explicitly set the memoized compute state to `valid = false` on initialization. I.e., expect that the element could have been manipulated / interrogated between construction and initialization.

theengineear:
This handles a nuanced edge case where internal, computed properties can be incorrect if they were interrogated between host construction and initialization.

For example:

```javascript
// This will trigger construction (assuming the element is defined).
const element = document.createElement('my-element');

// Do things after construction, before initialization.

// This will trigger initialization (again, assuming definition).
document.body.append(element);

// Do things after initialization while connected.
```

This change set should guarantee that after initialization, during connection — things will be correct. We cannot make such guarantees when not connected due to lack of eventing around attribute callbacks.

Closes #143.

### DIFF
--- a/test/test-computed-properties.js
+++ b/test/test-computed-properties.js
@@ -213,6 +213,19 @@ it('does correct NaN checking', () => {
   assert(el.count === count);
 });
 
+it('resets compute validity on initialization to catch upgrade edge cases with internal, computed properties', () => {
+  const el = document.createElement('test-element');
+  el.setAttribute('a', '1');
+  el.setAttribute('b', '2');
+  assert(el.a === undefined);
+  assert(el.b === undefined);
+  assert(Number.isNaN(el.internal.c));
+  document.body.append(el);
+  assert(el.a === 1);
+  assert(el.b === 2);
+  assert(el.internal.c === 3);
+});
+
 it('cannot be written to from host', () => {
   const el = document.createElement('test-element');
   document.body.append(el);
--- a/x-element.js
+++ b/x-element.js
@@ -646,7 +646,7 @@ export default class XElement extends HTMLElement {
 
   static #initializeHost(host) {
     const hostInfo = XElement.#hosts.get(host);
-    const { initialized, invalidProperties } = hostInfo;
+    const { computeMap, initialized, invalidProperties } = hostInfo;
     if (initialized === false) {
       XElement.#upgradeOwnProperties(host);
       // Only reflect attributes when the element is connected.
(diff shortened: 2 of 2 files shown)


test/test-computed-properties.js
```diff
@@ -213,6 +213,19 @@ it('does correct NaN checking', () => {
   assert(el.count === count);
 });
 
+it('resets compute validity on initialization to catch upgrade edge cases with internal, computed properties', () => {
```
theengineear, line 216: Here's the straightforward test case. Honestly, this is the most convincing argument IMO — it's clear that this should work as this test suggests. I.e., `1 + 2` should indeed equal `3` 😂

theengineear:
FYI @klebba — Going to get this change in since I'm confident on it and I'd like to include it in a project I'm working on. I left detailed code comments for you — we can always revert / change if you're opposed 👌

PR #148, opened 2023-08-02
Deserialize `""` to `NaN` for `Number` type.

### LINKED ISSUE #147: Deserializing empty attributes when the property type definition is Number results in "0"
Is it a bug?

Consider:

```
<x-foo baz></x-foo>
```

Where `baz` is defined as a Number type property; the value of `baz` will be coerced to `Number 0`. This is because, in Javascript, `Number('') === 0`.

theengineear:
It would always be counter-intuitive to declare `<my-element number>` in HTML markup and introspect the element to find `element.number === 0`.

In other words, while it’s true that in JS `Number('') === 0`, this is not sensible behavior when deserializing values in markup into typed objects in JS.

This is mostly due to the concept of so-called “boolean attributes”. As a custom element author, you don’t want to coerce a boolean attribute to a `Number`. In otherwords, when dealing with numbers, the markup `<my-element number>` should _not_ equal `<my-element number="0">`.

While this change makes a strong opinion, part of the goal of `x-element` is to reduce obvious boiler plate. Since it seems incredibly unlikely that any author would actually _desire_ the `Number('') >> 0` behavior, taking a stance here seems more helpful than hurtful.

Additionally, we will always, conceptually, have a _default_ deserializer. In the future, if needed, we could allow authors to declare their own deserializers / serializers within property declarations.

Closes #147.

### DIFF
--- a/test/test-basic-properties.js
+++ b/test/test-basic-properties.js
@@ -250,6 +250,19 @@ it('inheritance is considered in type checking', () => {
   assert(el.objectProperty === array, 'property was set');
 });
 
+it('numeric properties deserialize "" (empty) to "NaN"', () => {
+  const el = document.createElement('test-element');
+  el.setAttribute('numeric-property', '0');
+  document.body.append(el);
+  assert(el.numericProperty === 0, '"0" was coerced to 0');
+  el.setAttribute('numeric-property', '');
+  assert(Number.isNaN(el.numericProperty), '"" was coerced to NaN');
+  el.setAttribute('numeric-property', ' ');
+  assert(Number.isNaN(el.numericProperty), '" " was coerced to NaN');
+  el.setAttribute('numeric-property', '      ');
+  assert(Number.isNaN(el.numericProperty), '"      " was coerced to NaN');
+});
+
 it('cannot set to known properties', () => {
   class BadTestElement extends XElement {
     static get properties() {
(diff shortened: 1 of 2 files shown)


test/test-basic-properties.js
```diff
@@ -250,6 +250,19 @@ it('inheritance is considered in type checking', () => {
   assert(el.objectProperty === array, 'property was set');
 });
 
+it('numeric properties deserialize "" (empty) to "NaN"', () => {
```
theengineear, line 253: The spec locks down our new opinions. All of `''`, `' '`, `'    '` should be coerced to `NaN`. Note that no other tests were depending on the prior `Number('') >> 0` behavior.

theengineear:
@klebba — I really wrestled with myself over whether this opinion makes sense or not in `x-element`. Ultimately, I think because we have to opine on _some_ default, we may as well choose the one we think is best. We can always introduce `serialize` and `deserialize` functions as declarations in property blocks:

```
number: {
  type: Number,
  reflect: true,
  serialize: value => /* custom serialization routine to run on reflection */,
  deserialize: string => /* custom deserialization routine to run on sync */,
}
```

klebba:
Thanks @theengineear — it's worth noting that the impetus for this change was a UI bug in which a null value was mistakenly represented as "0" and subsequently depicted as a Unix epoch 1970-01-01 timestamp by a date formatter (or some such thing, I can't recall the details) — in other words, not a default behavior anyone actually wants.

klebba (approved):
Nice!

PR #150, opened 2023-08-18
Bump dev dependencies.

theengineear:
Just doing some periodic updates to the dependencies here to keep things fresh. Things seem to run fine with the new updates.

Note that the versions of node need to also be bumped in our GitHub actions files.


theengineear:
Just updating dev-side things in here. Going to go ahead and get this merged.

PR #155, opened 2023-12-07
Uses ESLint flat config format

klebba:
@theengineear FYI


theengineear (approved):
Thanks for the update!

Issue #153, opened 2023-11-20
Unpkg links should incorporate a version


klebba:
By default Unpkg will auto select the latest major version of the requested lib, as a result the lit-html demos are broken now that Lit 3 has been published there. See for example:

https://import.dog/@netflix/x-element/demo/lit-html/

See: https://github.com/Netflix/x-element/blob/main/demo/lit-html/base-element.js

```
https://unpkg.com/lit-html/directives/async-append.js?module
```

Becomes:
```
https://unpkg.com/lit-html@2/directives/async-append.js?module
```
Etc.

PR #156, opened 2024-02-01
Update import paths for networked demos.

theengineear:
In a some places, we demo loading 3rd party dependencies over-the-wire. The latest versions of these dependencies on `unpkg` require a slightly different url. Just a minor change to update those.


klebba:
Nice, thanks!

klebba:
@theengineear what is your take on #153 ?

theengineear:
Yep, we can do that as well 👌 — I was just getting this fixed quickly.

PR #159, opened 2024-02-02
Updates dev dependencies

klebba:
Aiming to address:

* https://github.com/Netflix/x-element/security/dependabot/16

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,9 +27,9 @@
   ],
   "dependencies": {},
   "devDependencies": {
-    "eslint": "^8.55.0",
+    "eslint": "^8.56.0",
     "http-server": "^14.1.1",
-    "puppeteer": "^21.0.3",
-    "tap-parser": "^13.0.1"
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -27,9 +27,9 @@
   ],
   "dependencies": {},
   "devDependencies": {
-    "eslint": "^8.55.0",
+    "eslint": "^8.56.0",
     "http-server": "^14.1.1",
-    "puppeteer": "^21.0.3",
-    "tap-parser": "^13.0.1"
+    "puppeteer": "^21.10.0",
+    "tap-parser": "^15.3.1"
```
klebba, line 33: Although the major version bumped I was still able to run `yarn test` without issue

theengineear: Seems like it’s all `TAP Version 14` still (from `x-test` to `tap-parser` — so no major concerns from me).

theengineear (approved):
👍

PR #165, opened 2024-02-02
Fixes server pathing

klebba:
Serves the root directory; now uses `<meta>` refresh to determine which page to start with

### DIFF
--- a/demo/lit-html/index.html
+++ b/demo/lit-html/index.html
@@ -1,6 +1,7 @@
 <!doctype html>
 <html>
   <head>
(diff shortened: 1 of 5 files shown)


demo/lit-html/index.html
```diff
@@ -1,6 +1,7 @@
 <!doctype html>
 <html>
   <head>
+    <meta charset="UTF-8">
```
klebba, line 4: oopsie. easy to forget that this really matters when using emojis!

theengineear:
Thanks!!

PR #166, opened 2024-02-02
Improve style sheet adoption ergonomics.

### LINKED ISSUE #52: Leverage the pattern of "Constructable Stylesheet Objects" for shared CSS
The general spec lives here: https://wicg.github.io/construct-stylesheets/#proposed-solution

Here's a Google article on it: https://developers.google.com/web/updates/2019/02/constructable-stylesheets

Here's how LitElement incorporates it: https://lit-element.polymer-project.org/guide/styles#static-styles
Something to keep an eye on!

Here's the motivation section from that spec for reference:

>Most web components uses Shadow DOM. For a style sheet to take effect within the Shadow DOM, it currently must be specified using a style element within each shadow root. As a web page may contain tens of thousands of web components, this can easily have a large time and memory cost if user agents force the style sheet rules to be parsed and stored once for every style element. However, the duplications are actually not needed as the web components will most likely use the same styling, perhaps one for each component library.
>
>Some user agents might attempt to optimize by sharing internal style sheet representations across different instances of the style element. However, component libraries may use JavaScript to modify the style sheet rules, which will thwart style sheet sharing and have large costs in performance and memory.

Food for thought:

```
// Helper method to create a stylesheet to be included via adoptedStylesheets.
static createStyleSheet(string) {
  const styleSheet = new CSSStyleSheet();
  styleSheet.replaceSync(string);
  return styleSheet;
}
```

theengineear:
One common pattern for element authors (now that import attributes enable folks to import css files directly) is to adopt imported style sheets into a shadow root at custom element initialization time.

This adds one static getters — `styles`. The goal is to make shadow root initialization as _declarative_ as possible. Note that we still expose `createRenderRoot` so we don’t get in the way if folks need to do something more advanced (it’s also still the only way to forgo the creation of a shadow root at all).

Closes #52.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- New support for static `styles` getter for `adoptedStyleSheets` ergonomics (#52).
+
+### Fixed
+
+- The `map` function now works with properties / attributes bound across template contexts (#179).
+
 ## [1.0.0] - 2024-02-29
 
 ### Added
--- a/test/test-render-root.js
+++ b/test/test-render-root.js
@@ -1,7 +1,7 @@
 import XElement from '../x-element.js';
 import { assert, it } from './x-test.js';
 
-class TestElement extends XElement {
+class TestElement1 extends XElement {
   static createRenderRoot(host) {
     return host;
   }
@@ -11,21 +11,21 @@ class TestElement extends XElement {
     };
   }
 }
-customElements.define('test-element', TestElement);
-
+customElements.define('test-element-1', TestElement1);
 
 it('test render root was respected', () => {
-  const el = document.createElement('test-element');
+  const el = document.createElement('test-element-1');
   document.body.append(el);
   assert(el.shadowRoot === null);
   assert(el.textContent === `I'm not in a shadow root.`);
+  el.remove();
 });
 
 it('errors are thrown in for creating a bad render root', () => {
   class BadElement extends XElement {
     static createRenderRoot() {}
   }
-  customElements.define('test-element-1', BadElement);
+  customElements.define('test-element-2', BadElement);
   let passed = false;
   let message = 'no error was thrown';
   try {
--- a/test/test-styles.js
+++ b/test/test-styles.js
@@ -1,0 +1,79 @@
+import { assert, it } from './x-test.js';
+import styleSheet from './test-styles.css.js';
+import XElement from '../x-element.js';
+
+class TestElement1 extends XElement {
+  static count = 0;
+  static get styles() {
+    TestElement1.count++;
+    return [styleSheet];
+  }
+  static template(html) {
+    return () => {
+      return html``;
+    };
+  }
+}
+customElements.define('test-element-1', TestElement1);
+
+it('provided style sheets are adopted', () => {
+  const el = document.createElement('test-element-1');
+  document.body.append(el);
+  const boundingClientRect = el.getBoundingClientRect();
+  assert(boundingClientRect.width === 100);
+  assert(boundingClientRect.height === 100);
+  el.remove();
+});
+
+it('should only get styles _once_ per constructor', () => {
+  for (let iii = 0; iii < 10; iii++) {
+    // No matter how many times you do this, styles must only be accessed once.
+    const el = document.createElement('test-element-1');
+    document.body.append(el);
+    const boundingClientRect = el.getBoundingClientRect();
+    assert(boundingClientRect.width === 100);
+    assert(boundingClientRect.height === 100);
+    el.remove();
+    assert(TestElement1.count === 1);
+  }
+});
+
+it('errors are thrown when providing styles without a shadow root', () => {
+  class BadElement extends XElement {
+    static get styles() { return [styleSheet]; }
+    static createRenderRoot(host) { return host; }
+  }
+  customElements.define('test-element-2', BadElement);
+  let passed = false;
+  let message = 'no error was thrown';
+  try {
+    new BadElement();
+  } catch (error) {
+    const expected = 'Unexpected "styles" declared without a shadow root.';
+    message = error.message;
+    passed = error.message === expected;
+  }
+  assert(passed, message);
+});
+
+it('errors are thrown when styles already exist on shadow root.', () => {
+  class BadElement extends XElement {
+    static get styles() { return [styleSheet]; }
+    static createRenderRoot(host) {
+      host.attachShadow({ mode: 'open' });
+      host.shadowRoot.adoptedStyleSheets = [styleSheet];
+      return host.shadowRoot;
+    }
+  }
+  customElements.define('test-element-3', BadElement);
+  let passed = false;
+  let message = 'no error was thrown';
+  try {
+    new BadElement();
+  } catch (error) {
+    const expected = 'Unexpected "styles" declared when preexisting "adoptedStyleSheets" exist.';
+    message = error.message;
+    passed = error.message === expected;
+  }
+  assert(passed, message);
+});
--- a/x-element.js
+++ b/x-element.js
@@ -11,7 +11,8 @@ export default class XElement extends HTMLElement {
     return TemplateEngine.interface;
   }
 
-  /** Configured templating engine. Defaults to "defaultTemplateEngine".
+  /**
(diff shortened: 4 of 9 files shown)


x-element.js
```diff
@@ -21,6 +22,24 @@ export default class XElement extends HTMLElement {
     return XElement.defaultTemplateEngine;
   }
 
+  /**
+   * Declare an initialization object for the host’s shadow root.
+   * See https://developer.mozilla.org/en-US/docs/Web/API/Element/attachShadow.
+   */
+  static get shadowRootInit() {
```
theengineear, line 29: We’ve talked about pulling this chunk of configuration into a declarative getter before — seems much more reasonable now that we have _multiple_ things to configure when creating the render root.

x-element.js
```diff
@@ -21,6 +22,24 @@ export default class XElement extends HTMLElement {
...
+    return { mode: 'open' };
+  }
+
+  /**
+   * Declare an array of CSSSTyleSheet objects to adopt on the shadow root.
+   * Note that a CSSStyleSheet object is the type returned when importing a
+   * stylesheet file via import attributes. This has no effect if you are using
+   * the “host” as your render root (versus attaching a shadow root).
+   */
+  static get styleSheets() {
```
theengineear, line 39: This is the crux of the change. We can fuss over the name here. I personally like `cssStyleSheets`, `styleSheets`, or `styles` as the options. But, I’m willing to be convinced otherwise 😸

x-element.js
```diff
@@ -68,7 +87,9 @@ export default class XElement extends HTMLElement {
    * E.g., setup focus delegation or return host instead of host.shadowRoot.
    */
   static createRenderRoot(host) {
```
theengineear, line 89: With these changes — it’s far less likely that you will need to override the `createRenderRoot` method. IMO, that’s a good thing. As an author, I prefer to be able to declare things _in isolation_ and let the library cobble everything together.

test/test-render-root.js
```diff
@@ -11,21 +11,55 @@ class TestElement extends XElement {
     };
   }
 }
-customElements.define('test-element', TestElement);
+customElements.define('test-element-1', TestElement1);
+
+class TestElement2 extends XElement {
+  static get styleSheets() {
+    // TODO: Replace with direct import of css file when better-supported in
```
theengineear, line 18: You’ll have to use your imagination a bit until import attributes are better-supported. This is the pattern we’d be using though.

theengineear:
@klebba — This is more-or-less what I figured we’d do here. It’s pretty straightforward, just a way to declare things in isolation as an ergonomic benefit.

theengineear:
Note… if we _did_ want to get rid of `createRenderRoot`. We could consider a flag like this:

```
static get useShadowRoot() {
  return true;
}
```

We would have to do _something_ more to allow the use-case where folks will want to create a custom element _without_ a shadow root by using the `host` itself as the render root.

test/test-render-root.js
theengineear: We no longer need to use our imagination :^)

We might be able to update these to import the CSS in directly. Assuming all our testing clients are using Chrome, that should probably work!.

x-element.js
```diff
@@ -21,6 +22,24 @@ export default class XElement extends HTMLElement {
...
+  /**
+   * Declare an initialization object for the host’s shadow root.
+   * See https://developer.mozilla.org/en-US/docs/Web/API/Element/attachShadow.
+   */
+  static get shadowRootInit() {
+    return { mode: 'open' };
+  }
+
+  /**
+   * Declare an array of CSSSTyleSheet objects to adopt on the shadow root.
```
klebba, line 34: nit: typo `STyle`

x-element.js
```diff
@@ -68,7 +87,9 @@ export default class XElement extends HTMLElement {
    * E.g., setup focus delegation or return host instead of host.shadowRoot.
    */
   static createRenderRoot(host) {
-    return host.attachShadow({ mode: 'open' });
+    const shadowRoot = host.attachShadow(this.shadowRootInit);
```
klebba, line 90: I would recommend against the `shadowRootInit` concept — developers should simply override `createRenderRoot` to customize the render root init

x-element.js
```diff
@@ -68,7 +87,9 @@ export default class XElement extends HTMLElement {
...
+    shadowRoot.adoptedStyleSheets = this.styleSheets;
```
klebba, line 91: I recommend moving this out of `createRenderRoot` and instead moving it to the constructor — it would throw an error if there is no `shadowRoot` on the host

x-element.js
klebba:
```
import MyStylesAreTheBest from './path-to.css' with { type: 'css' };
import YourStyles from 'https://your.style/sheet.css' with { type: 'css' };

static get styles() {
  
  const rules = `:host { display: block }`;
  const inlineStyles = new CSSStyleSheet();
  inlineStyles.replaceSync(rules);

  return [
    MyStylesAreTheBest,
    YourStyles,
    inlineStyles,
  ];
}
```

test/test-render-root.js
theengineear: … except for ESLint doesn’t dig this yet. So close!

x-element.js
```diff
@@ -534,7 +544,18 @@ export default class XElement extends HTMLElement {
...
+      if (renderRoot === host.shadowRoot) {
+        if (renderRoot.adoptedStyleSheets.length === 0) {
+          renderRoot.adoptedStyleSheets = styles;
+        } else {
+          throw new Error('Unexpected "styles" declared when preexisting "adoptedStyleSheets" exist.');
+        }
+      } else {
+        throw new Error('Unexpected "styles" declared without a shadow root.');
+      }
+    }
```
theengineear, line 558: Kind of a lot of lines dedicated to error-handling… but so it goes 🤷

CHANGELOG.md
```diff
@@ -6,6 +6,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- New support for static `styles` getter for `adoptedStyleSheets` ergonomics (#52).
+
+### Fixed
+
+- The `map` function now works with properties / attributes bound across template contexts (#179).
```
theengineear, line 15: Just adding this in the `[Unreleased]` section — forgot to request this earlier.

test/test-styles.js
```diff
@@ -0,0 +1,79 @@
...
+it('provided style sheets are adopted', () => {
+  const el = document.createElement('test-element-1');
+  document.body.append(el);
+  const boundingClientRect = el.getBoundingClientRect();
+  assert(boundingClientRect.width === 100);
+  assert(boundingClientRect.height === 100);
+  el.remove();
+});
+
+it('should only get styles _once_ per constructor', () => {
```
theengineear, line 28: Importantly, this function is called _once_ since it’s a static member, we follow our typical pattern — assume it can be set up once during static analysis and re-used for all instances.

x-element.js
```diff
@@ -21,6 +22,15 @@ export default class XElement extends HTMLElement {
     return XElement.defaultTemplateEngine;
   }
 
+  /**
+   * Declare an array of CSSStyleSheet objects to adopt on the shadow root.
+   * Note that a CSSStyleSheet object is the type returned when importing a
```
klebba, line 27: nit: an example syntax might be more useful here

x-element.js
```diff
@@ -197,7 +207,7 @@ export default class XElement extends HTMLElement {
 
   // Called once per class — kicked off from "static get observedAttributes".
   static #analyzeConstructor(constructor) {
-    const { properties, listeners } = constructor;
+    const { styles, properties, listeners } = constructor;
```
klebba, line 210: nit: alpha sort

x-element.js
```diff
@@ -534,7 +544,18 @@ export default class XElement extends HTMLElement {
     const computeMap = new Map();
     const observeMap = new Map();
     const defaultMap = new Map();
-    const { propertyMap } = XElement.#constructors.get(host.constructor);
+    const { styles, propertyMap } = XElement.#constructors.get(host.constructor);
+    if (styles.length > 0) {
```
klebba, line 548: nit: `if (styles.length)`

klebba (approved):

klebba:
Looks awesome, nice work!

x-element.js
theengineear: I’ll add some additional commentary into the changes to use JSDoc 👌

Issue #169, opened 2024-02-13
Simplify template error message once error causes are better-supported.


theengineear:
See #128 for the original motivation here. Just opening this ticket to simplify some of our error logging to keep things as tidy as possible.

Current:

```js
  render() {
    const { template, properties, renderRoot, render } = XElement.#hosts.get(this);
    const result = template(properties, this);
    try {
      render(renderRoot, result);
    } catch (error) {
      const pathString = XElement.#toPathString(this);
      const message = `${error.message} — Invalid template for "${this.constructor.name}" at path "${pathString}"`;
      throw new Error(message, { cause: error });
    }
  }
```

Proposed:

```js
  render() {
    const { template, properties, renderRoot, render } = XElement.#hosts.get(this);
    const result = template(properties, this);
    try {
      render(renderRoot, result);
    } catch (error) {
      const pathString = XElement.#toPathString(this);
      const message = `Invalid template for "${this.constructor.name}" at path "${pathString}"`;
      throw new Error(message, { cause: error });
    }
  }
```

☝️ — See the difference? We ought to not have to literally repeat / embed the original error in our own messaging.

theengineear:
Note that this is currently supported in Firefox. Here’s a link to the Chrome [bug](https://issues.chromium.org/issues/40182832).

<img width="1292" alt="Screenshot 2024-02-13 at 1 15 37 PM" src="https://github.com/Netflix/x-element/assets/6611546/35077a41-3410-4a9e-a341-7b2b76c634d9">

theengineear:
This seems to work in all browsers now. BUT — we may want to consider that the `cause` there may or may not itself have an _error_ associated. I.e., an error `cause` could be the original error, or it could be _any_ compound object… I’m actually surprised / disappointed to learn this. It feels like a bad abstraction to me 🤷 

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/cause

theengineear:
We’ve made the related changes here. Closing.

PR #170, opened 2024-02-13
Leverage error “cause” for wrapping render errors.

### LINKED ISSUE #128: Use error cause for template errors.
The [error cause](https://github.com/tc39/proposal-error-cause) TC39 proposal was accepted in 2022 and is [available](https://caniuse.com/mdn-javascript_builtins_error_cause) in Chrome, Safari, and Firefox.

We currently use the following to inject helpful context into otherwise out-of-context errors [in render](https://github.com/Netflix/x-element/blob/main/x-element.js#L150-L160):

```javascript
  render() {
    const { template, properties, renderRoot } = XElement.#hosts.get(this);
    const result = template(properties, this);
    try {
      render(result, renderRoot);
    } catch (error) {
      const pathString = XElement.#toPathString(this);
      error.message = `${error.message} — Invalid template for "${this.constructor.name}" at path "${pathString}"`;
      throw error;
    }
  }
```

Instead, we could simply throw our own error and set the `cause` appropriately:

```javascript
  render() {
    const { template, properties, renderRoot } = XElement.#hosts.get(this);
    const result = template(properties, this);
    try {
      render(result, renderRoot);
    } catch (error) {
      const pathString = XElement.#toPathString(this);
      const message = `${error.message} — Invalid template for "${this.constructor.name}" at path "${pathString}"`;
      throw new Error(message, { cause: error });
    }
  }
```

This built-in "error chaining" allows you to link an arbitrary number of errors together without needing to mutate anything (or make up a bespoke pattern). Let's do it.

theengineear:
Because some template engines can throw _immutable_ errors, it’s safer and better to use an “error cause” to _chain_ the old error to our own versus trying to mutate the original.

Closes #128.

### DIFF
--- a/x-element.js
+++ b/x-element.js
@@ -123,8 +123,8 @@ export default class XElement extends HTMLElement {
       render(renderRoot, result);
     } catch (error) {
       const pathString = XElement.#toPathString(this);
-      error.message = `${error.message} — Invalid template for "${this.constructor.name}" at path "${pathString}"`;
-      throw error;
+      const message = `${error.message} — Invalid template for "${this.constructor.name}" at path "${pathString}"`;
+      throw new Error(message, { cause: error });
     }
   }
 


x-element.js
```diff
@@ -123,8 +123,8 @@ export default class XElement extends HTMLElement {
       render(renderRoot, result);
     } catch (error) {
       const pathString = XElement.#toPathString(this);
-      error.message = `${error.message} — Invalid template for "${this.constructor.name}" at path "${pathString}"`;
-      throw error;
+      const message = `${error.message} — Invalid template for "${this.constructor.name}" at path "${pathString}"`;
```
theengineear, line 126: Note, for now, we continue to embed the original error  message. Eventually, we can simply remove it once the tooling catches up — see #169.

theengineear:
FYI @klebba — super small change to leverage error “cause”. It’s a slightly safer way of wrapping errors.

PR #171, opened 2024-02-28
Add “deno.json” to configure support for JSR.

theengineear:
Just following directions at https://jsr.io/@netflix/x-element/publish.


theengineear:
Note that this got added as part of a separate effort so closing this down as a duplicate. For completeness — note that this _does_ publish to JSR as part of our release pipeline via GitHub Actions 👌

PR #172, opened 2024-02-28
Update GitHub actions (versions / ergonomics).

theengineear:
This change set bumps the referenced / reusable GitHub actions to their latest versions, bumps node versions, and enables reuse of a “test” wf.


.github/workflows/publish-to-npm.yaml
```diff
@@ -0,0 +1,23 @@
...
+jobs:
+  main:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v3
+        with:
+          node-version: '20.x'
+          registry-url: 'https://registry.npmjs.org'
+      - run: npm ci
```
theengineear, line 20: In general — I’ve updated things to use [clean install](https://docs.npmjs.com/cli/v9/commands/npm-ci) so that we install from `package-lock.json` for consistency.

.github/workflows/publish-to-npm.yaml
```diff
@@ -0,0 +1,23 @@
...
+          description: 'A token to enable this workflow to publish to npm.'
+          required: true
+jobs:
+  main:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v3
+        with:
+          node-version: '20.x'
```
theengineear, line 18: Node version is bumped to `20.x` in all these configuration files.

.github/workflows/publish.yaml
```diff
@@ -0,0 +1,12 @@
+# Run tests and publish to all registries if successful.
+name: Publish
+on:
+  release:
+    types: [published]
```
theengineear, line 5: Updated the release type trigger to `published` as that appears to be the current recommendation.

.github/workflows/test.yaml
```diff
@@ -1,17 +1,16 @@
-name: Push
-on: [push]
+# Test flow triggered by other workflows.
```
theengineear, line 1: Also, I’ve added a general change to pull out our workflows into reusable chunks. The GitHub UI sort of rewards you with a better UX for doing this (e.g., better graphical representation of what’s happening).

theengineear:
FYI @klebba — I’m just modernizing our GH actions stuff as we flesh out this ecosystem. I’m going to get this merged in and then will start testing the deployment flows.

.github/workflows/publish-to-npm.yaml
```diff
@@ -0,0 +1,23 @@
+# Publish package to npm whenever a release is pushed.
+#  See https://docs.github.com/en/actions/publishing-packages/publishing-nodejs-packages
+name: Publish to npm
+on:
+  workflow_call:
+    inputs:
+      secrets:
```
theengineear, line 7: Going to try and remove this — I believe that if we use `secrets: inherit`, you don’t need to explicitly declare inputs like that.

.github/workflows/test.yaml
```diff
@@ -1,17 +1,16 @@
...
-  test:
+  main:
     runs-on: ubuntu-latest
-    permissions:
-      contents: read
     steps:
-      - uses: actions/checkout@v2
-      - uses: actions/setup-node@v2
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v3
```
theengineear, line 9: Bumped all these as part of the node 16 >> 20 upgrades in GH actions.

PR #174, opened 2024-02-29
Publish package to JSR in addition to npm.

theengineear:


bump.sh
```diff
@@ -0,0 +1,45 @@
...
+if [ -z "${1}" ]
+then
+  echo "Pass in a version string (e.g., major, minor, 1.0.0-rc.58, 2.3.4, etc.)."
+  exit 0
+fi
+
+# Bump version in package.json using npm version itself. This ensures that we
+#  stay anchored to first-class tooling. We pass “--git-tag-version=false” to
+#  prevent the command from (1) committing changes and (2) creating a tag.
+prefixed_version="$(npm version --git-tag-version=false "${1}")"
```
theengineear, line 22: This is critical – note that we are 100% passing _exactly_ the value to the real `npm version` command here. This means that running `npm version patch` (or some such) will _just work_. It also means that if you try to provide the _same_ version or something, it’ll fail with a human-readable error message.

PR #177, opened 2024-04-09
Adopts shared linter rules

klebba:
One new violation:

```
~/x-element.js
  1328:53  warning  Function declared in a loop contains unsafe references to variable(s) 'node'  no-loop-func

✖ 1 problem (0 errors, 1 warning)
```

### DIFF
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -1,38 +1,11 @@
 import globals from 'globals';
-import js from '@eslint/js';
-
-const common = {
-  rules: {
-    ...js.configs.recommended.rules,
-    'comma-dangle': ['warn', {
-      arrays: 'always-multiline',
-      objects: 'always-multiline',
-      imports: 'always-multiline',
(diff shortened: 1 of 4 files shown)


klebba:
@theengineear how do you recommend I address the `no-loop-func` violation? I could ignore it or rework it (could use your help with that)

theengineear:
Can you just change that `find` function from:

```js
    const find = path => {
      let node = content;
      for (const index of path) {
        node = Template.#setIfMissing(lookup, node, () => node.childNodes)[index];
      }
      return node;
    };
```

to

```js
    const find = path => {
      let node = content;
      for (const index of path) {
        const ref = node;
        node = Template.#setIfMissing(lookup, node, () => ref.childNodes)[index];
      }
      return node;
    };
```

klebba:
Thanks for the explainer! Fixt.

theengineear (approved):
LGTM!

Issue #179, opened 2024-04-26
Issue with built-in template repeat


klebba:
TODO: needs a working repro case.

What's wrong:

The repeater does not re-render when `data` changes, only when the list of `nodes` is updated:

```
${repeat(nodes, node => node.id, node => html`
  <my-node
    id="${node.id}"
    .metadata="${node.metadata}"
    .data="${data[node.id]}">
  </my-node>
`)}
```

Workaround:

Coalesce `nodes` with `data` into a new list to feed the repeater (ensure new list reference is always generated)

Using version 1.0

klebba:
FYI @theengineear I'll try to build a working example eventually, just bookmarking this for later

klebba:
👋🏻 @Kelketek thank you for taking up this issue. Just wanted to share an FYI that we recognize there are no contributing guidelines for you to follow and opened #186 to document a few thoughts. As you are the first notable contributor to the project your feedback is welcome on the guidelines as well. Thanks!

Kelketek:
Thanks, @klebba ! I'll take a look :)

klebba:
Repro case:

```
<!doctype html>
<html>
  <head>
    <meta charset="UTF-8">
    <title>demo #179 issue</title>
  </head>
  <body>
    <demo-179></demo-179>
    <script type="module">
    import XElement from "https://deno.land/x/element@v1.0.0/x-element.js";

    class Demo179 extends XElement {
      static get properties()	{
        return {
          items: {
            type: Array,
            default: () => [
              { id: 'a' },
              { id: 'b' },
              { id: 'c' }
            ],
          },
          lookup: {
            type: Object,
            default: () => {
              return {
                a: 'foo',
                b: 'bar',
                c: 'baz',
              };
            }
          },
        }
      }

      static template(html, { repeat }) {
        return ({ items, lookup }) => {
          return html`
            <div>
              <ul>
              ${
                repeat(items, item => item.id, item => {
                  return html`<li id="${ item.id }">${ lookup?.[item.id] }</li>`;
                })
              }
              </ul>
              <pre>${JSON.stringify(lookup, null, 2)}</pre>
            </div>
          `;
        };
      }
    }

    customElements.define('demo-179', Demo179);
    </script>
    <script>
    	const demo = document.querySelector('demo-179');
      let flip = 0;
      setInterval(() => {
        if (flip % 2 === 0) {
          demo.lookup = {
            a: 'sizzle',
            b: 'pop',
            c: 'buzz',
[...]

PR #180, opened 2024-04-26
Adds docs about templating, minor housekeeping

klebba:
Covers templating in more detail. Creates `doc` directory.


doc/TEMPLATES.md
```diff
@@ -0,0 +1,208 @@
...
+
+Add a static template function in your `x-element` definition in order to leverage automagical DOM generation and data binding:
+
+```
+static template(html, { repeat }) {
+  return ({ options, selectedId }) => {
+    return html`
+      <select name="my-options">
+        ${repeat(options, option => option.id, option => html`
+          <option value="${option.value}" ?selected="${option.id === selectedId}">
```
klebba, line 13: TODO: does this example work? #179 tracks a potential problem

doc/TEMPLATES.md
```diff
@@ -0,0 +1,213 @@
...
+
+// text bindings assign the value to the content of the node
+el.textContent = foo;
+
+// content bindings create and append text to the node
+el.appendChild(document.createTextNode(foo))
+```
+
+### Important note on serialization during data binding:
+Values assigned to DOM attributes are always serizalized using `toString()` during assignment. To help you avoid `[object Object]` surprises, properties defined using `x-element` allow you to specify their anticipated type. Properties with scalar types `String`, `Number`, and `Boolean` may be bound to attributes using native serialization. Attempting to bind non-scalar types to attributes will result in an `x-element` error message.
```
Kelketek, line 53:
```suggestion
Values assigned to DOM attributes are always serialized using `toString()` during assignment. To help you avoid `[object Object]` surprises, properties defined using `x-element` allow you to specify their anticipated type. Properties with scalar types `String`, `Number`, and `Boolean` may be bound to attributes using native serialization. Attempting to bind non-scalar types to attributes will result in an `x-element` error message.
```

klebba: Oh, nice catch! Thank you!

doc/TESTING.md
theengineear, line 1: The way I typically test locally is to `npm start`, then, visit the `/test` page in the browser. Can we cover that case in here as well?

package.json
```diff
@@ -27,9 +27,9 @@
   ],
   "devDependencies": {
     "@netflix/eslint-config": "^3.0.0",
-    "eslint": "^9.0.0",
-    "puppeteer": "^22.0.0",
-    "tap-parser": "^15.3.1"
+    "eslint": "^9.8.0",
+    "puppeteer": "^23.0.2",
+    "tap-parser": "^18.0.0"
```
theengineear, line 32: Hrm, they sorta bundled all the `tapjs` stuff into a single changelog. It’s much harder to tell exactly what changed related to the `tap-parser` library. Anyways, it should still be working with TAP 14 output, so that should still interop with `x-test`.

doc/TEMPLATES.md
klebba: it works now — but we should use `map`

doc/TEMPLATES.md
```diff
@@ -0,0 +1,211 @@
...
+
+| Type                | Example                |
+| :------------------ | :--------------------- |
+| attribute           | `<div foo="${bar}">`   |
+| attribute (boolean) | `<div ?foo="${bar}">`  |
+| property            | `<div .foo="${bar}">`  |
+| text                | `<div>${foo}</div>`    |
+| content             | `${foo}`               |
+
+Equivalent to:
```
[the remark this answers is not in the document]
klebba: how about 
> emulates

doc/TEMPLATES.md
```diff
@@ -0,0 +1,211 @@
...
+
+## Summary
+
+Features distributed as custom elements are framework and library agnostic. Thus custom elements can integrate with [any modern framework](https://custom-elements-everywhere.com/). By using native ShadowDOM encapsulation developers can choose the manner in which they manage the DOM while avoiding the risk of vendor lock-in.
+
+Key concepts repeated:
+
+* Custom elements are not a framework (native feature)
+* Custom elements provide DOM, JS and CSS encapsulation (native feature)
+* Developers can choose a framework to manage the DOM within their custom element
```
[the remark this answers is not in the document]
klebba: thats good feedback — shall we defer to a future revision?

theengineear (approved):
LGTM!

PR #182, opened 2024-07-20
Documentation: Clarify that computed properties are lazy.

### LINKED ISSUE #178: Lazily computed properties can cause confusion
A developer may attempt to access a lazily computed property value and be confused as to why it is not evaluated. Computed properties will only be evaluated when they are observed, used as input in downstream property computation, or included in the template function. Can we produce a warning in the console when an attempt is made to read from a still-lazy property?

Kelketek:
Small documentation change-- clarifies that computed properties are lazily evaluated. Resolves #178.

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -151,6 +151,10 @@ callbacks are memoized and are expected to be pure.
 If the given function is defined on the constructor, the context (`this`) is
 guaranteed to be the constructor when called later.
 
+**Computed properties are lazily evaluated.** If a computed property is not
+rendered in a template function or accessed programmatically, its code will
+not run.
+
 ## Observed properties
 
 Create an observed property by providing an `observe` function. Whenever an


theengineear (approved):
Thanks for the update @Kelketek — makes sense to call that out explicitly 👌

klebba:
+1 — thanks @Kelketek — out of curiosity could you share how you came across this project/what you are using it for? Would love to hear more — thanks!

Kelketek:
@klebba I just started browsing Netflix's frontend repositories, curious what you all were working on. I found this interesting since previous to this point I've been using tools like React and Vue, and I hadn't seen much in terms of serious use of Web Components.

Diving into this project was a good way to learn about Web Components-- I'd not used them before. I'm hoping to have a PR for you soon to resolve #179 as well, or at least greatly mitigate it between a code tweak and additional documentation. I should hopefully have that for you by tomorrow.

klebba:
Thanks for the insight @Kelketek and it's nice to hear that this project was helpful for you!

PR #185, opened 2024-08-09
Update server.js

Osintedx:
the If statement shouldn't be the only statement in the else block on line 43, its bad code practice.

It was now repaired!


klebba:
@Osintedx the proposed diff appears to refactor the whole contents of `server.js` — some of the modifications would cause behavior changes. Let's work out what it is you hope to address in the issue thread before you spend more time here. Thanks.

PR #188, opened 2024-10-09
Improve internal “template engine”.

### LINKED ISSUE #187: Consolidate / improve internal template engine logic.
Because the entirety of the templating engine is opaque to integrators, the internal interfaces have never really been scrutinized.

Fortunately, _because_ it’s all internal, we can take this step-by-step and incrementally improve upon what we have without needing to do major / minor version bumps.

Some things to TODO:
* Improve class / method names to match expectations.
* Reduce the number of _layers_ (ideally, we eventually just have a `TemplateEngine` class).
* Create some documentation describing how things work at a high-level.

theengineear:
Nearly all of the template engine details are kept internal — this allowed us to ship a first pass of the code without needed high-levels of scrutiny. However, we can (and will!) revisit this internal code to make it more readable / self-documenting.

Closes #187.

### DIFF
--- a/x-element.js
+++ b/x-element.js
@@ -899,16 +899,38 @@ export default class XElement extends HTMLElement {
   static #prototypeInterface = new Set(Object.getOwnPropertyNames(XElement.prototype));
 }
 
-/** Wrapper to document public interface to the default templating engine. */
+/** Internal implementation details for template engine. */
 class TemplateEngine {
+  static #UNSET = Symbol('unset'); // Ensures a unique, initial comparison.
+  static #ATTRIBUTE_PREFIXES = {
+    attribute: 'x-element-attribute-',
+    boolean: 'x-element-boolean-',
+    property: 'x-element-property-',
+  };
+  static #CONTENT_PREFIX = 'x-element-content-';
+  static #CONTENT_REGEX = new RegExp(`${TemplateEngine.#CONTENT_PREFIX}(\\d+)`);
+  static #OPEN = /<[a-z][a-z0-9-]*(?=\s)/g;
+  static #STEP = /\s+[a-z][a-z0-9-]*(?=[\s>])|\s+[a-z][a-zA-Z0-9-]*="[^"]*"/y;
+  static #ATTRIBUTE = /\s+(\??([a-z][a-zA-Z0-9-]*))="$/y;
+  static #PROPERTY = /\s+\.([a-z][a-zA-Z0-9_]*)="$/y;
+  static #CLOSE = />/g;
+
   static #interface = null;
+  static #stateMap = new WeakMap(); // Maps nodes to internal state.
+  static #analysisMap = new WeakMap(); // Maps strings to cached computations.
+  static #resultMap = new WeakMap(); // Maps references to results.
+  static #updaterMap = new WeakMap(); // Maps references to updaters.
 
   /**
    * Declare HTML markup to be interpolated.
+   *
    *   html`<div attr="${obj.attr}" .prop="${obj.prop}">${obj.content}</div>`;
    */
   static html(strings, ...values) {
-    return Template.html(strings, ...values);
+    const reference = TemplateEngine.#createReference();
+    const result = TemplateEngine.#createResult('html', strings, values);
+    TemplateEngine.#resultMap.set(reference, result);
+    return reference;
   }
 
   /**
@@ -917,7 +939,10 @@ class TemplateEngine {
    *   svg`<circle r="${obj.r}" cx="${obj.cx}" cy="${obj.cy}"></div>`;
    */
   static svg(strings, ...values) {
-    return Template.svg(strings, ...values);
+    const reference = TemplateEngine.#createReference();
+    const result = TemplateEngine.#createResult('svg', strings, values);
+    TemplateEngine.#resultMap.set(reference, result);
+    return reference;
   }
 
   /**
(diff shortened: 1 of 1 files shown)


TEMPLATE_ENGINE.md
```diff
@@ -0,0 +1,126 @@
+# Template Engine
```
theengineear, line 1: When I was revisiting some of the code — I found myself wishing a document like this existed. Additionally, I think that having to describe some of this stuff _in prose_ has helped me refactor / rename the actual code to be more self-documenting.

This does get into some nitty-gritty implementation details, but I think it might be worth keeping around for developer perusal.

x-element.js
```diff
@@ -1662,47 +1688,3 @@ class Template {
     }
   }
 }
-
-/** Internal implementation details for template results. */
-class TemplateResult {
```
theengineear, line 1667: Pretty much all of the diff here is a flattening / renaming of concepts. I found myself squinting to explain why we have a function called `template`, a class called the `TemplateEngine`, a class called `Template` and a class called `TemplateResult`…

_Now_, we can (roughly) say the following:

> The “template” function returns a “template result” by leveraging a “template engine”.

^^ As the saying goes the wrong abstraction is worse than no abstraction at all! Rather than have all those extra classes, we’re en route to just having `TemplateEngine`.

theengineear:
@klebba — I was motivated by some recent activity in here to start making some improvements to the internal `TemplateEngine` code. I’m certain that there’s more to be done here, but I think this is a decent iteration.

The TL;DR is that I’ve removed some poor abstraction layers in favor of moving towards a single class called `TemplateEngine`.

TEMPLATE_ENGINE.md
```diff
@@ -0,0 +1,126 @@
...
+
+## Concepts
+
+The goal of `XElement` is to make it easier for developers to author Custom
```
klebba, line 5: nit: One goal

TEMPLATE_ENGINE.md
klebba, line 1: (Advanced topic)
Goal: dispel FUD about template engines — they are simple
Goal: Explain how the thing really works

- "Templating" in web development is essentially a way for developers to provide a specification for how they want their living DOM to behave throughout changing/variable conditions of their data
- Templating happens outside and inside an encapsulated element — they are completely discrete from custom elements. a template engine DOES NOT KNOW that `<div>` is different from `<my-fancy-button>`
- Markdown is a template engine
- Engines ultimately generate DOM nodes and make imperative calls to modify them
- Engines manage efficient creation and deletion of DOM nodes
- Engines manage efficient updates to DOM nodes
- Engines manage efficient reuse of DOM nodes

Philosophy:
- one important aspirational goal for the x-element built-in template engine is that developers should not need to learn a new DSL — they only need to author idiomatic javascript. there are a few exceptions we call 'directives' and 'binding syntax' which a developer must learn to leverage the tool, however we aim to keep this list TINY (and shrinking over time whenever possible)

note: add (DRAFT) to the heading of this doc?

x-element.js
```diff
@@ -1294,100 +1226,113 @@ class Template {
...
     }
     let iii = 0;
     for (const childNode of node.childNodes) {
-      items.push(...Template.#evaluate(childNode, [...path, iii++]));
+      initialDirections.push(...TemplateEngine.#getInitialDirections(childNode, [...path, iii++]));
+    }
+    return initialDirections;
+  }
+
+  static #fillInHtml(type, strings) {
```
klebba, line 1261: `populateHTML` ?

x-element.js
```diff
@@ -8,7 +8,7 @@ export default class XElement extends HTMLElement {
 
   /** Default templating engine. Use "templateEngine" to override. */
   static get defaultTemplateEngine() {
-    return TemplateEngine.interface;
+    return TemplateEngineWrapper.interface;
```
klebba, line 11: nit: can we eliminate `TemplateEngineWrapper`?

TEMPLATE_ENGINE.md
```diff
@@ -0,0 +1,126 @@
...
+Elements. A key part of that functionality is enabling a declarative way to turn
+attributes and properties into DOM — that’s the role of the “template engine”.
+
+So, a `template` function like this is using the `html` method from a
+“template engine” to create a “template result” — which will be “rendered” into
+some “container” (either the “host” element or the host’s “shadow root”).
+
+```js
```
klebba, line 13:
```
class FooElement extends XElement {
  static get properties() {
    return {
      bar: { type: String },
    };
  }
  static template(html) {
    return ({ bar }) => {
      return html`
        <div>${bar}</div>
      `;
    };
  }
}
customElements.define('foo-element', FooElement);
```

is equivalent to

```
host.shadowRoot.querySelector('div').textContent = bar;
```

TEMPLATE_ENGINE.md
theengineear: As discussed — I’ll create a separate PR for the documentation portion.

x-element.js
theengineear: Going to leave this as-is in here for now.

x-element.js
theengineear: I reworked some of this to remove this wrapper. As mentioned, the main point was internal ergonomics around the `repeat` shim. I just created another internal `#mapOrRepeat` function to handle it so that the external interface remains intact.

theengineear:
@klebba — Going to get this merged in. I’m only making internal changes so no interface changes here. I also punted the documentation portion out and I can open up a new PR to address that.

klebba:
Nice work!

TEMPLATE_ENGINE.md
theengineear: Here’s that other PR — https://github.com/Netflix/x-element/pull/193

PR #189, opened 2024-10-09
Generated type declarations

theengineear:
Generate “x-element.d.ts” from JSDoc comments.

Previously, the “x-element.d.ts” file was hand-curated, which was prone
to falling out of date.

Because the TypeScript team manages JSDoc, it’s fairly straightforward
to properly “type” your JS with JSDoc and output declarations which can
be used by folks writing TypeScript — all of that without needing to
actually author TypeScript ourselves.†

Additionally, tools like JSR will build basic documentation for
libraries which author JSDoc comments, so some wins there.

Finally, to ensure best practices for our JSDocs, we enable some
recommended rules from the eslint plugin.

† One goal of “x-element” is to be _highly_ portable, it’s why no
  dependencies exist and why the “x-element.js” file can be used
  _verbatim_ without a build step.


x-element.d.ts
```diff
@@ -1,69 +1,239 @@
-export class XElement extends HTMLElement {
```
theengineear, line 1: Here’s a good example — this should have said `export default class XElement extends HTMLElement {` (see the `default`). Easy to miss since we often don’t depend on this module. Generating it from well-considered JSDoc “types” feels like a better way to ensure this stays accurate.

x-element.d.ts
```diff
@@ -1,69 +1,239 @@
...
+     *     }
+     *   }
+     * @param {Function} html
+     * @param {{[key: string]: Function}} engine
+     * @returns {templateCallback}
+     */
+    static template(html: Function, engine: {
+        [key: string]: Function;
+    }): (properties: object, host: HTMLElement) => any;
+    static "__#1@#analyzeConstructor"(constructor: any): void;
```
theengineear, line 131: … This, is annoying. TypeScript does want to track private fields for [reasons], so as far as I can tell, you cannot remove this stuff.

package.json
```diff
@@ -3,11 +3,12 @@
   "description": "A dead simple starting point for custom elements.",
   "version": "1.0.0",
   "license": "Apache-2.0",
-  "repository": "https://github.com/Netflix/x-element",
+  "repository": "github:Netflix/x-element",
```
theengineear, line 6: I went through some of the npm documentation again and tried to further “modernize” this `package.json` — worth further scrutiny for sure though.

package.json
```diff
@@ -17,19 +18,23 @@
     "lint": "eslint --max-warnings=0 .",
     "lint-fix": "eslint --fix .",
     "test": "node test.js | tap-parser -l",
+    "build": "tsc",
```
theengineear, line 21: Still need to figure out how we want to include this output…

tsconfig.tsbuildinfo
```diff
@@ -0,0 +1 @@
+{"root":["./x-element.js"],"version":"5.6.3"}
```
theengineear, line 1: Not sure what this is all about. Can we `.gitignore` this?

theengineear:
Another step I’d like to take is publishing this branch to JSR as throw-away exercise. I’d be curious to see if it significantly improves the experience there.

tsconfig.tsbuildinfo
Kelketek: @theengineear You **could** merge the contents into the tsconfig.json. This file is helpful for splitting the project into multiple [project references](https://www.typescriptlang.org/docs/handbook/project-references.html). However, TypeScript also just allows you to [define these references](https://www.typescriptlang.org/tsconfig/#references) in the `tsconfig.json` if you need them anyway. It just supports allowing you using a separate config file, for which this file name [is the default](https://www.typescriptlang.org/tsconfig/#tsBuildInfoFile). I don't know why someone would choose to maintain a separate file from ts-config-- that probably makes more sense from the perspective of very large projects.

Anyway, since x-element is intended to be a single file import, I don't imagine you'll need this unless you're planning to add accessory projects to the repo. I'd just delete it for now. The version property doesn't appear to be an officially supported property for tsconfig, and you're already using the `files` attribute instead of `root`.

package.json
Kelketek: @theengineear Do you need to? If you're using JSDoc, TypeScript will import the file as typed.

tsconfig.tsbuildinfo
theengineear: Thanks for the detailed response! I expected to be the one answering my own question there 😅 

I’ll ignore the file 👌

x-element.js
```diff
@@ -899,110 +975,122 @@ export default class XElement extends HTMLElement {
   static #prototypeInterface = new Set(Object.getOwnPropertyNames(XElement.prototype));
 }
 
-/** Wrapper to document public interface to the default templating engine. */
-class TemplateEngine {
+// TODO: If / when we remove `repeat` in favor of `map` — we can probably delete
```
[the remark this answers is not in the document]
theengineear: The only thing we really need to do is have the next release be a `2.x` release so that we’re following semantic versioning. It’s not much of a hassle to drag along `repeat` until we’re actually ready to make that change though 👌 

_This way_, we can continue shipping incremental changes in `1.x` without having to manage multiple, major versions or anything.

package.json
theengineear: Since the entire interface is really a single class — I do think it makes sense to keep this checked into source as a sibling of the file itself. It allows us to scrutinize any changes that could happen there.

That said… that’s just my current opinion there. It’s completely subject to change if it ends up being overly cumbersome!

.github/workflows/test.yaml
```diff
@@ -12,6 +12,8 @@ jobs:
           registry-url: 'https://registry.npmjs.org'
       - run: npm ci
       - run: npm run lint
+      - run: npm run type
+      - run: git diff --exit-code
```
theengineear, line 16: Rather than add a “build” step. We now have a CI step that will fail if a diff is generated from running `npm run type`. This is closer to the flow I _think_ we want. I.e., rather than need to run a build to use the types… assert that the types were properly generated before allowing the merge.

tsconfig.json
```diff
@@ -0,0 +1,19 @@
...
+    "target": "ES2023",
+    "allowJs": true,
+    "checkJs": true,
+    "declaration": true,
+    "emitDeclarationOnly": true,
+    // Even though the docs suggest that “declarationMap” is a way to go from
+    //  generated-js back to ts… it also seems to be the case that it will help
+    //  IDEs get to the right _js_ when using types. See
+    //  https://www.typescriptlang.org/docs/handbook/declaration-files/dts-from-js.html
+    "declarationMap": true,
```
theengineear, line 15: After re-reviewing this. I _think_ it’s actually most correct to emit a source map.

1. If you’re using the JS on it’s own — the source map is inert (since our JS is running as-is).
2. If you’re using the TS declaration file, the source map will hint IDEs to route you back to the original JS.

This was tricky to wrap my head around at first since I typically think about these source mappings as a way to go from the _generated JS_ back to the _original TS_ — but, it seems like you can leverage them to hint IDEs to do the opposite as well.

package.json
```diff
@@ -17,19 +18,24 @@
...
     "/x-element.js",
+    "/x-element.d.ts",
+    "/x-element.d.ts.map",
     "/demo",
     "/test",
     "/etc"
   ],
   "devDependencies": {
     "@netflix/eslint-config": "^3.0.0",
+    "eslint-plugin-jsdoc": "^50.3.1",
```
theengineear, line 34: After using this for a little while — I feel like this is a pretty good solution. Open to other suggestions though. FWIW, I tried using TypeScript to generate errors here and I kept feeling like it wasn’t doing what I wanted it to do — could be that I had it set up wrong though!

x-element.d.ts
```diff
@@ -1,69 +1,240 @@
...
+    /**
+     * Observe callback.
+     * @callback observeCallback
+     * @param {HTMLElement} host
+     * @param {any} value
+     * @param {any} oldValue
+     */
+    /**
+     * A property value.
+     * @typedef {object} Property
```
theengineear, line 34: I’ll be curious to see how tools like JSR auto-doc this stuff!

theengineear:
^^ Just tested the new CI control. It does indeed fail the build if a diff is detected and it prints the diff.

Screenshot 2024-10-15 at 10 41 38 PM

theengineear:
Fiddling around a bit with uploading to JSR — the docs are good in some ways (and rough around the edges in other ways). But, increasingly, it’d be cool to be generating our docs from the actual code itself.

Screenshot 2024-10-15 at 11 01 49 PM

x-element.js
```diff
@@ -1,112 +1,176 @@
...
-   *         type: Number,
-   *         input: ['property1'],
-   *         compute: this.computeProperty2,
-   *         reflect: true,
-   *         observe: this.observeProperty2,
-   *         default: 0,
-   *       }
-   *     };
-   *   }
+   * ```js
```
theengineear, line 56: JSDoc also has an `@example` tag… but this feels more natural to write code blocks like this and it appears to be supported by JSR.

package.json
```diff
@@ -17,19 +18,24 @@
     "lint": "eslint --max-warnings=0 .",
     "lint-fix": "eslint --fix .",
     "test": "node test.js | tap-parser -l",
+    "type": "tsc",
     "bump": "./bump.sh"
   },
   "files": [
     "/x-element.js",
+    "/x-element.d.ts",
+    "/x-element.d.ts.map",
```
klebba, line 27: do we need to generate a source map?

deno.json
```diff
@@ -1,5 +1,12 @@
 {
   "name": "@netflix/x-element",
   "version": "1.0.0",
-  "exports": "./x-element.js"
+  "exports": {
```
klebba, line 4: this seems wrong?

klebba (dismissed):
awesome!

theengineear:
FYI, I think JSR is gonna wrongly flag this package as “slow” until https://github.com/jsr-io/jsr/issues/494 is resolved. I _think_ this should all work eventually though 👌

deno.json
theengineear: JSR’s documentation is not great here. I’ve added back the other files, which causes JSR to flag us as “slow” and throw warnings. I _think_ that’s ok for now though.

I’m going to follow along on https://github.com/jsr-io/jsr/issues/494 — hopefully things will get better once our JSDoc-style path forward here is accepted.

theengineear:
Here’s a screen grab of an `rc` publish to JSR I ran locally (a more zoomed out view). Again, not perfect, but definitely getting there.

Screenshot 2024-10-18 at 3 26 23 PM

theengineear:
Going to get this merged in — I’ll play around with some `rc` releases to make sure everything still works as expected 👌

PR #196, opened 2024-10-22
Use non-registered symbols as weak map keys.

### LINKED ISSUE #192: Use non-registered symbols as weak map keys.
The [proposal](https://github.com/tc39/proposal-symbols-as-weakmap-keys) to allow non-registered symbols (e.g., `Symbol('foo')`) as keys for weak maps was accepted for 2023. Seems sensible to pivot over to using something like `Symbol('__result__')` for the opaque references we pass back from our internal template engine.

Currently, `x-element` creates a new _null object_ (`Object.create(null)`), which feels dated now that symbols can be used 👌

theengineear:
This is now baseline as of 2023 so we ought to be able to use it. We use symbols to prevent internal abstractions from being leaked into code authored by integrators — for that, we just need an opaque token. But, because weak maps previously needed to be keyed with an _object_, we just created a dummy object as our key.

Now, we can take the more modern approach of keying with a symbol.

Closes #192.

### DIFF
--- a/x-element.js
+++ b/x-element.js
@@ -1010,7 +1010,7 @@ export default class XElement extends HTMLElement {
 
 /** Internal implementation details for template engine. */
 class TemplateEngine {
-  static #UNSET = Symbol('unset'); // Ensures a unique, initial comparison.
+  static #UNSET = Symbol(); // Ensures a unique, initial comparison.
   static #ATTRIBUTE_PREFIXES = {
     attribute: 'x-element-attribute-',
     boolean: 'x-element-boolean-',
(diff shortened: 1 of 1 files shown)


x-element.js
```diff
@@ -1756,10 +1756,6 @@ class TemplateEngine {
     }
   }
 
-  static #createReference() {
-    return Object.create(null);
```
theengineear, line 1760: We’re simply trading `Object.create(null)` for `Symbol()` in this code. To me, calling `Symbol()` (note, not `Symbol.for()`) has better semantics — it’s pretty clear that we just want a guaranteed-unique reference there.

x-element.js
```diff
@@ -1010,7 +1010,7 @@ export default class XElement extends HTMLElement {
 
 /** Internal implementation details for template engine. */
 class TemplateEngine {
-  static #UNSET = Symbol('unset'); // Ensures a unique, initial comparison.
```
theengineear, line 1013: I decided that the labels for the Symbols isn’t actually additive. I could potentially be convinced otherwise, but the code context is always sufficient to understand the meaning.

x-element.js
```diff
@@ -1040,10 +1040,10 @@ class TemplateEngine {
    * @returns {any}
    */
   static html(strings, ...values) {
-    const reference = TemplateEngine.#createReference();
+    const symbol = Symbol();
```
theengineear, line 1043: Again, we could have say like `Symbol('html')` or `Symbol('result')` here — but I really don’t think that’s additive. If anything, it makes it seem like the string has meaning when it really doesn’t for our use-case.

theengineear:
@klebba — This one is pretty minor. Just wanted to keep you in the loop. Let me know if you have any questions on it!

klebba (approved):
Noice!

PR #198, opened 2024-11-08
Perform rendering in a fully  depth-first manner.

### LINKED ISSUE #197: Default template engine connects children before binding updates.
I ran into this sneaky situation recently when the _initial_ value bound to some child element really mattered (i.e., eventual consistency was not good enough).

I believe the crux of the problem boils down to the fact that we insert newly minted DOM into our `container` _before_ we bind updates into that DOM. This causes the new elements to get their `connectedCallback` called _before_ we actually set initial values.

In my particular case — this caused _multiple_ observations, which caused inadvertent animations!

Here is a dummy repro…

```js
class InnerElement extends XElement {
  static get properties() {
    return {
      foo: {
        observe: (host, value) => console.log(value),
        default: false,
      },
    };
  }
}
customElements.define('inner-element', InnerElement);

class OuterElement extends XElement {
  static template(html) {
    return () => html`<inner-element .foo="${true}"></inner-element>`;
  }
}
customElements.define('outer-element', OuterElement);

const el = document.createElement('outer-element');
document.body.append(el);
// false
// true
```

☝️ — We see a `false` get logged first! That‘s the bug. It’s never ever the case that the inner `foo` property should see any other value than `true`… and yet…

theengineear:
Currently, we inject dom into our container and then update bindings. This causes all child “connectedCallback” calls to fire _before_ we bind property / attribute / etc. updates.

That’s problematic because elements have no good way to observe initial deltas in the render tree. Said another way…

We used to do this:
 1. Create new nodes and inject new nodes into our container.
 2. Bind value updates to new nodes.

Now we do this:
 1. Create new nodes.
 2. Bind value updates to new nodes.
 3. Inject new nodes into our container.

Closes #197.

### DIFF
--- a/x-element.js
+++ b/x-element.js
@@ -1074,12 +1074,14 @@ class TemplateEngine {
       const result = TemplateEngine.#resultMap.get(resultReference);
       if (TemplateEngine.#cannotReuseResult(state.result, result)) {
         TemplateEngine.#removeWithin(container);
-        state.result = result;
+        TemplateEngine.#ready(result);
+        TemplateEngine.#commit(result);
         TemplateEngine.#inject(result, container);
+        state.result = result;
       } else {
         TemplateEngine.#assign(state.result, result);
+        TemplateEngine.#commit(state.result);
       }
-      TemplateEngine.#commit(state.result);
     } else {
       TemplateEngine.#clearObject(state);
       TemplateEngine.#removeWithin(container);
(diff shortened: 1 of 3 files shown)


x-element.js
```diff
@@ -1074,12 +1074,14 @@ class TemplateEngine {
       const result = TemplateEngine.#resultMap.get(resultReference);
       if (TemplateEngine.#cannotReuseResult(state.result, result)) {
         TemplateEngine.#removeWithin(container);
-        state.result = result;
+        TemplateEngine.#ready(result);
+        TemplateEngine.#commit(result);
         TemplateEngine.#inject(result, container);
```
theengineear, line 1079: The above three lines sum up the change neatly. Before, we readied / injected in one line… then committed the update. Now, we ready, commit, and _lastly_ inject.

Yes, in the future we will change all these internal method names… not now though.

theengineear:
@klebba — As a back-burner task, I try to take a moment every once and a while to kick the tires on the default template engine. Here’s an interesting bug I found / fixed today.

PR #202, opened 2024-11-10
Improve template issue forensics.

### LINKED ISSUE #201: Improve template issue forensics.
**Proposed**

```
Invalid template for "AppPlatformSearchPopularItem" / <app-platform-search-popular-item> at path "app-platform-search-popular-item < div[id="starred"] < div[id="popular-links"][hide] < div[id="container"][data-page="search-tables"][data-initializing][data-deferring][data-status="loading"] < app-platform-search < x-switch[id="pages"][valid][value="app-platform-search"] < div[id="body"] < div[id="application"][data-show-menu][data-show-scratchpads][style="--app-scrollbar-height: 0px;"] < div[id="viewport"][selected="appPlatformSearch"] < app-view[show-new-scratchpad-button] < app-root < body < html[lang="en"]".

Caused by: Error: Found invalid template at line 52 within string `` at ``.
```

It’s better because:
1. The line number where the issue occurred in the template is printed. Worth its weight in gold.
2. The custom element name is printed with angle brackets _in addition to_ the constructor name — this is incredibly helpful because you can easily search for like `customElements.define` usage with the provided string literal. We cannot really point to the actual file path, but this is pretty much as good.
3. The repetitive error chaining is no longer necessary (error causes get printed the way you want in the console now).

**Current**

```
Found invalid template string "" at "". — Invalid template for "AppPlatformSearchPopularItem" at path "app-platform-search-popular-item < div[id="starred"] < div[id="popular-links"][hide] < div[id="container"][data-page="search-tables"][data-initializing][data-deferring][data-status="loading"] < app-platform-search < x-switch[id="pages"][valid][value="app-platform-search"] < div[id="body"] < div[id="application"][data-show-menu][data-show-scratchpads][style="--app-scrollbar-height: 0px;"] < div[id="viewport"][selected="appPlatformSearch"] < app-view[show-new-scratchpad-button] < app-root < body < html[lang="en"]"

Caused by: Error: Found invalid template string "" at "".
```

**Context**

I was actually unable to quickly detect where the problem was in the template… so much so that I hacked in the proposed output which could tell me _which line_ the issue occurred on. Of all the changes here, that’s incredibly helpful. In this case, it was caused by a weird attribute binding like this `foo="${bar}${baz}"` (hence the cryptic error with the empty strings).

theengineear:
The major changes can be boiled down into the following additions:
* Print the registered custom element tag name in the error.
* Print the approximate template line number in the error.†

† Note that this isn’t the _file line number_, just a naive count of
  the number of newlines encountered so far in the given template.

Closes #201


test/test-initialization-errors.js
```diff
@@ -142,7 +142,7 @@ it('errors are thrown in connectedCallback when template result fails to render
   try {
     el.connectedCallback();
   } catch (error) {
-    const expected = ' — Invalid template for "TestElement" at path "test-element-4[id="testing"][class="foo bar"][boolean][variation="primary"]"';
+    const expected = 'Invalid template for "TestElement" / <test-element-4> at path "test-element-4[id="testing"][class="foo bar"][boolean][variation="primary"]".';
```
theengineear, line 145: Seeing something like `<test-element-4>` is often more useful than the actual class name for finding the file-in-question. I honestly don’t know why we hadn’t done this yet 🤔

test/test-template-engine.js
```diff
@@ -884,7 +884,7 @@ describe('rendering errors', () => {
       } catch (e) {
         error = e;
       }
-      assert(error?.message === `Found invalid template string "<div id="target" .notOk=" at " .notOk=".`, error.message);
+      assert(error?.message === `Found invalid template at line 4 within string \`\n\n\n<div id="target" .notOk=\` at \` .notOk=\`.`, error.message);
```
theengineear, line 887: The `at line 4` text here is wildly helpful . It makes it exceedingly easy to track down template issues. IMO, if we’re gonna be strict about templates, we should strive to be increasingly specific about what went wrong and where.

theengineear:
@klebba — I was kicking the tires again on migrating to the default template engine. These changes make the process _much_ more efficient.

test/test-template-engine.js
```diff
@@ -856,12 +857,14 @@ describe('rendering errors', () => {
       } catch (e) {
         error = e;
       }
-      assert(error?.message === `Found invalid template string "<div id="target" not-ok=" at " not-ok=".`, error.message);
+      assert(error?.message === `Found invalid template on or after line 1 in substring \`<div id="target" not-ok=\`. Failed to parse \` not-ok=\`.`, error.message);
       container.remove();
     });
 
+    // Note that this is _still_ line “1” because it begins with a newline. This
+    //  is what authors will expect as the “first line”.
```
theengineear, line 865: This is pretty nuanced, but I want the following to _both_ conceptually report the “first line”:

```js
html`<div .oops='${whoops}'></div>`;
```

```js
html`
  <div .oops='${whoops}'></div>
`;
```

… i.e., when the first line is _literally just a newline and nothing else_, ignore it, it’s not what I would consider the _first line_. The rational here is that if / when you could copy this text into a separate html file (not in our lifetimes), it would match the line numbering there.

x-element.js
```diff
@@ -1385,7 +1387,14 @@ class TemplateEngine {
...
             state.index = 1; // Accounts for an expected quote character next.
           } else {
-            throw new Error(`Found invalid template string "${string}" at "${string.slice(state.index)}".`);
+            // We try and intuit what authors consider line “1”.
+            const handled = [...strings.slice(0, iii), string.slice(0, state.index)].join('');
+            const lineCount = handled.startsWith('\n')
+              ? handled.split('\n').length - 1
+              : handled.split('\n').length;
+            // We say “on or after” because interpolated JS could add lines we
+            //  cannot know about or account for.
```
theengineear, line 1396: Note that you _can_ write templates like this, hence the “on or after” phrase:

```js
html`
  <div .ok="${







                           'i can haz lines'


                                                }" .oops='${whoops}'>
  </div>
`;
```

This is gonna report like “on or after line 1” and there’re really not much we can do about that!

test/test-template-engine.js
klebba: Interesting — IMO stripping the newlines is more confusing because the actual code is:
```

  <div .oops='${whoops}'></div>

```

I might prefer to keep this precise vs smart, but I do not feel strongly about it

x-element.js
```diff
@@ -1385,7 +1387,14 @@ class TemplateEngine {
             string = string.slice(0, -3 - name.length) + `${TemplateEngine.#ATTRIBUTE_PREFIXES.property}${iii}="${name}`;
             state.index = 1; // Accounts for an expected quote character next.
           } else {
-            throw new Error(`Found invalid template string "${string}" at "${string.slice(state.index)}".`);
+            // We try and intuit what authors consider line “1”.
```
klebba, line 1390: Seeing the fork here does make me think that we should avoid exceptions like this — if the user chooses to author templates in a certain manner that will have implications for line numbering the library should remain agnostic

test/test-template-engine.js
theengineear: > stripping the newlines

🤔 — Hmm, definitely not stripping newlines! More like conditionally deleting the initial newline. I think you’d agree that in either case, the html file would look like this, right? I.e., the markup there _is on line one_. I can remove the fork though. Not worth arguing 🤙 

```html
<div></div>
```

theengineear:
@klebba — Gonna get this one merged in based on our conversation earlier 🤙

Issue #204, opened 2024-11-12
Clarify null and undefined serialization binding behavior in docs


klebba:
There are some nuances regarding binding `null` and `undefined` binding behavior in XElement — philosophically we want to leverage (and not augment) the browser's behavior.

Add detailed info here: https://github.com/Netflix/x-element/blob/main/doc/TEMPLATES.md

Maybe update the philosophy: "Please do not make me learn anything proprietary about the framework unless it is absolutely necessary!"

Consider:

```javascript
// uses element.textContent = binding
html`<div>${null}</div>`
// -> <div></div>
html`<div>${undefined}</div>`
// -> <div></div>


// uses element.innerHTML = binding (unsafe!)
html`<div>${unsafeHTML(null)}</div>`
// -> <div></div>
html`<div>${unsafeHTML(undefined)}</div>`
// -> <div>undefined</div>


// uses element.setAttribute('name', binding)
html`<div test-null="${null}"></div>`
// -> <div test-null="null"></div>
html`<div test-undefined="${undefined}"></div>`
// -> <div test-undefined="undefined"></div>


// uses element.setAttribute('name', binding)
html`<div test-null="${null ?? ''}"></div>`
// -> <div test-null></div>
html`<div test-undefined="${undefined ?? ''}"></div>`
// -> <div test-undefined></div>


// uses binding ? element.setAttribute('name', '') : element.removeAttribute('name') (boolean attribute behavior)
html`<div ?test-null="${null}"></div>`
// -> <div></div>
html`<div ?test-undefined="${undefined}"></div>`
// -> <div></div>


// magic helper (undefined)
/*
 if (binding === undefined) {
   element.removeAttribute('name')
 } else {
   element.setAttribute('name', binding)
}
*/
html`<div test-null="${ifDefined(null)}"></div>`
// -> <div test-null="null"></div>
html`<div test-undefined="${ifDefined(undefined)}"></div>`
// -> <div></div>


// magic helper (nullish)
/*
 if (binding === undefined || binding === null) {
   element.removeAttribute('name')
 } else {
   element.setAttribute('name', binding)
}
*/
html`<div test-null="${nullish(null)}"></div>`
// -> <div></div>
html`<div test-undefined="${nullish(undefined)}"></div>`
// -> <div></div>


// These cases throw an error
html`<div>${ifDefined(null)}</div>`
html`<div>${ifDefined(undefined)}</div>`
html`<div>${nullish(null)}</div>`
[...]

PR #205, opened 2024-11-12
Remove nullish coalescing for `textContent`.

theengineear:
The default browser behavior when setting `.textContent` to `null` or `undefined` is to ultimately set the content to `''`†. This is quite different from the behavior of `createTextNode` though… by splitting the text node creation into multiple steps we can consistently leverage `.textContent` and remove the need to `?? ''`.

† Note that the WHATWG spec only defines that behavior for `null`, but
  in practice all modern browsers seem to treat `undefined` similarly.

Here’s the [spec](https://dom.spec.whatwg.org/#dom-node-textcontent) for reference. This is related to #204.

### DIFF
--- a/test/test-basic-properties.js
+++ b/test/test-basic-properties.js
@@ -23,6 +23,9 @@ class TestElement extends XElement {
         type: String,
         default: null,
       },
+      undefinedProperty: {
+        type: String,
+      },
       typelessProperty: {},
       typelessPropertyWithCustomAttribute: {
         attribute: 'custom-attribute-typeless',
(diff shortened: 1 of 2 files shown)


test/test-basic-properties.js
```diff
@@ -60,6 +64,12 @@ it('renders an empty string in place of null value', () => {
   assert(el.shadowRoot.getElementById('nul').textContent === '');
 });
 
+it('renders an empty string in place of undefined value', () => {
```
theengineear, line 67: Note that this passed before, we’re just specifically testing it now.

x-element.js
```diff
@@ -1488,10 +1488,11 @@ class TemplateEngine {
         }
         const previousSibling = node.previousSibling;
         if (previousSibling === startNode) {
-          const textNode = document.createTextNode(value ?? '');
```
theengineear, line 1491: This was the only place we _really_ needed to coalesce. If you pass `null` or `undefined` here you get `'null'` or `'undefined'`.

theengineear:
FYI @klebba — Hopefully this one is pretty easy to review. Shouldn’t be too controversial.

theengineear:
@klebba — Gonna get this in to unblock myself on some other work.

PR #209, opened 2024-11-16
Adding a couple minor changes for 2.x interface.

theengineear:
Changes:
* Stricter handling of “textarea” binding. Previously we were overly lenient in how the binding worked which caused automagical behavior. For example, newlines were quietly ignored, which could lead to issues if you really _wanted_ newlines in a default value in a text area control and were actually trying to interpolate between lines. Now, it just throws for that case.
* Spec-compliant attribute traversal. Previously, we relied on the browser convention that attributes in a NamedNodeMap would be iterated over in a particular manner. The spec strictly indicates that this is not to be relied on — so we can oblige.
* The deprecated `plaintext` html tag is no longer considered. Use at your own peril…
* General improvements to inline documentation and naming of things.

### DIFF
--- a/test/test-template-engine.js
+++ b/test/test-template-engine.js
@@ -65,6 +65,75 @@ describe('html rendering', () => {
     container.remove();
   });
 
+  // Unlike a NodeList, a NamedNodeMap (i.e., “.attributes”) is not technically
+  //  ordered in any way. This test just confirms that the template engine logic
+  //  doesn’t get confused in any way post-parse.
+  it('renders elements with many attributes in a weird order', () => {
+    const getTemplate = ({
+      property1,
+      z99,
+      defined,
+      dataFoo,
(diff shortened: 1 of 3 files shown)


test/test-template-engine.js
```diff
@@ -1058,6 +1127,38 @@ describe('rendering errors', () => {
...
+      } catch (e) {
+        error = e;
+      }
+      assert(error?.message === `Only basic interpolation of "textarea" tags is allowed.`, error.message);
+      container.remove();
+    });
+
+    it('throws when attempting non-trivial interpolation of a title tag', () => {
+      const getTemplate = ({ content }) => {
+        return html`<title id="target">please ${content} no</title>`;
```
theengineear, line 1148: We used to allow this sort of thing! Definitely not what we want.

PR #213, opened 2024-11-19
Additional updates for “2.x” release.

theengineear:
Additional updates for “2.x” release.

Changes:
* Deprecate `unsafeHTML` and `usafeSVG`.
* Allow binding `DocumentFragment` as a value.
* Tag every line in the “CHANGELOG.md” with an issue ticket.
* Simplified formatting related to bindings in “TEMPLATES.md”.
* Emit deprecation warnings for soon-to-be-gone interfaces.
* Deprecate remaining updaters (e.g., “map”).
* Ditch unecessary “weak maps”.†
* Some performance improvements.

† The search set for the weak maps we’re using gets untenable since it’s
  a flat list of pointers for _all_ results. It feels like a reasonable
  concession to hang data off of a unique symbol key. These are
  non-enumerable (unless specifically trying to enumerate symbols),
  which feels internal enough. Additionally, returning the values passed
  in by the user doesn’t feel like much of a leaking abstraction.

### DIFF
--- a/doc/TEMPLATES.md
+++ b/doc/TEMPLATES.md
@@ -8,62 +8,33 @@ Add a static template function in your `x-element` definition in order to
 leverage automagical DOM generation and data binding:
 
 ```javascript
-static template(html, { map }) {
+static template(html) {
   return ({ options, selectedId }) => {
     return html`
       <select name="my-options">
-        ${map(options, option => option.id, option => html`
-          <option value="${option.value}" ?selected="${option.id === selectedId}">
-        `)}
+        ${options.map(option => [
+          option.id,
+          html`<option value="${option.value}" ?selected="${option.id === selectedId}">`,
+        ])}
       </select>
     `;
   };
 }
 ```
 
-The following binding types are supported:
+The following bindings are supported:
 
-| Type                | Example                                    |
-| :------------------ | :----------------------------------------- |
-| attribute           | `<span id="target" foo="${bar}"></span>`   |
-| attribute (boolean) | `<span id="target" ?foo="${bar}"></span>`  |
-| attribute (defined) | `<span id="target" ??foo="${bar}"></span>` |
-| property            | `<span id="target" .foo="${bar}"></span>`  |
-| content             | `<span id="target">${foo}</span>`          |
-
-Emulates:
-
-```javascript
-const el = document.createElement('div');
-el.attachShadow({ mode: 'open' });
-el.innerHTML = '<span id="target"></span>';
-const target = el.shadowRoot.getElementById('target');
-
-// attribute value bindings set the attribute value
-target.setAttribute('foo', bar);
-
-// attribute boolean bindings set the attribute to an empty string or remove
-target.setAttribute('foo', ''); // when bar is truthy
-target.removeAttribute('foo'); // when bar is falsy
-
-// attribute defined bindings set the attribute if the value is non-nullish
-target.setAttribute('foo', bar); // when bar is non-nullish
-target.removeAttribute('foo'); // when bar is nullish
-
-// property bindings assign the value to the property of the node
-target.foo = bar;
-
-// content bindings create text nodes for basic content
-const text = document.createTextNode('');
-text.textContent = foo;
-target.append(text);
-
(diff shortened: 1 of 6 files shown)


doc/TEMPLATES.md
```diff
@@ -21,49 +21,19 @@ static template(html, { map }) {
...
+| :------------------ | :--------------------------- | :------------------------------------------------------------ |
+| --                  | --                           | `const el = document.createElement('div');`                   |
+| attribute           | `<div foo="${bar}"></div>`   | `el.setAttribute('foo', bar);`                                |
+| attribute (boolean) | `<div ?foo="${bar}"></div>`  | `el.setAttribute('foo', ''); // if  “bar” is truthy`          |
+| --                  | --                           | `el.removeAttribute('foo'); // if  “bar” is falsy`            |
+| attribute (defined) | `<div ??foo="${bar}"></div>` | `el.setAttribute('foo', bar); // if  “bar” is non-nullish`    |
+| --                  | --                           | `el.removeAttribute('foo'); // if  “bar” is nullish`          |
+| property            | `<div .foo="${bar}"></div>`  | `el.foo = bar;`                                               |
+| content             | `<div>${foo}</div>`          | `el.append(document.createTextNode(foo)) // if “bar” is text` |
+| --                  | --                           | (see [content binding](#content-binding) for composition)     |
```
theengineear, line 36: Made the table a bit more wide so that the emulation code is proximal to the template code.

test/test-template-engine.js
```diff
@@ -564,7 +564,7 @@ describe('html updaters', () => {
 
   it('unsafe html', () => {
     const getTemplate = ({ content }) => {
-      return html`<div id="target">${unsafe(content, 'html')}</div>`;
+      return html`<div id="target">${unsafe(content)}</div>`;
```
theengineear, line 567: The `unsafe` updater gets much more simple.

test/test-template-engine.js
```diff
@@ -1476,28 +1451,10 @@ describe('rendering errors', () => {
 
 
   describe('unsafe', () => {
-    it('throws if used on an unexpected language', () => {
```
theengineear, line 1479: We don’t support `unsafe` svg for now. We may choose to add this back in the future in some backwards-compatible way.

theengineear:
FYI — @klebba.

x-element.js
```diff
@@ -1918,18 +1872,51 @@ class TemplateEngine {
...
   }
 
-  // Attach a document fragment into some container. Note that all the DOM in
-  //  the fragment will already have values correctly bound.
-  static #inject(result, node, options) {
+  // Inject a given result into a node for the first time. If we’ve never seen
+  //  the template “strings” before, we also have to generate html, parse it,
+  //  and find out binding targets. Then, we commit the values by iterating over
+  //  our targets. Finally, we actually attach our new DOM into our node.
+  static #inject(result, node, before) {
```
theengineear, line 1879: @klebba — Did some refactoring / renaming here since you called it out as confusing when you were looking things over yesterday. Hope it helps!

x-element.js
```diff
@@ -1855,16 +1812,13 @@ class TemplateEngine {
           if (TemplateEngine.#cannotReuseResult(state.result, result)) {
             TemplateEngine.#removeBetween(startNode, node);
             TemplateEngine.#clearObject(state);
-            TemplateEngine.#ready(result);
-            TemplateEngine.#commit(result);
-            TemplateEngine.#inject(result, node, { before: true });
+            TemplateEngine.#inject(result, node, true);
```
theengineear, line 1815: The `ready + commit + inject` are collapsed into a single `inject` method now.

x-element.js
```diff
@@ -1855,16 +1812,13 @@ class TemplateEngine {
...
             state.result = result;
           } else {
-            TemplateEngine.#assign(state.result, result);
-            TemplateEngine.#commit(state.result);
+            TemplateEngine.#update(state.result, result);
```
theengineear, line 1818: The `assign + commit` methods are collapsed into a single `update` method now.

PR #215, opened 2024-11-19
Final “2.x” interface.

theengineear:
This mainly deletes things we deprecated in “1.x” which enables a smooth transition from “1.x” >> “2.x”.


theengineear:
@klebba — You asked about a PR to show what the interface will look like. Here one is! This is just throwaway work since it’s pointing to a different PR, but I’ll keep recreating every time I make changes so that you can stay updated here.

theengineear:
The git diff shown in the `test-template-engine.js` file is not great… it’s really a straight deletion… bummer that it shows up like that 🤷

klebba:
> x-template contents are always valid HTML and this is important so that HTML linters can be used to evaluate them

^ can we incorporate this somewhere?

x-template.js
```diff
@@ -314,62 +252,21 @@ class TemplateEngine {
...
-    const update = TemplateEngine.#symbolToUpdate.get(value);
-    if (update) {
-      TemplateEngine.#throwIfDefinedError(TemplateEngine.#PROPERTY);
-    } else {
-      node[name] = value;
-    }
+    node[name] = value;
   }
 
-  // TODO: Future state here once “ifDefined” is gone.
```
theengineear, line 357: The future is now!

PR #225, opened 2024-11-25
Improve performance tests.

theengineear:
Changes:

* Attempt to test more iterations.
* Do better at right-sizing batches per animation frame.
* Only measure batches to get around security issues with timestamps.
* Show distribution of results for better relative analysis.

Screenshot 2024-11-25 at 9 53 32 AM

### DIFF
--- a/demo/performance/index.css
+++ b/demo/performance/index.css
@@ -2,17 +2,74 @@
   line-height: 20px;
 }
 
+.test + .test {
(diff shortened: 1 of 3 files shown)


theengineear:
I was playing around with some performance tuning over the weekend and found myself frustrated with the feedback / development cycles. I spent a moment yesterday to improve the way in which we report performance.

PR #227, opened 2024-12-05
Split out template engine into “x-template.js”.

theengineear:
Conceptually “XElement” and “XTemplate” do very different things. The former is all about defining _element interfaces_ and the latter is all about turning interpolated markup into dom.

Over time, even our PR diffs have highlited this separation of concerns. Often a PR _either_ changes XElement _or_ XTemplate — rarely both.


tsconfig.json
```diff
@@ -1,10 +1,15 @@
 {
   "files": [
     "./x-element.js",
+    "./x-template.js",
     "./etc/ready.js"
   ],
   "compilerOptions": {
     "target": "ES2023",
+    // Beyond just organization, a “bug” in how JSDocs works causes a cyclic
```
theengineear, line 9: This was a cute one. It’s not really a _bug_ from the perspective of the maintainers. Not terribly hard to workaround though — we just throw things into a `/ts` directory. Maybe we should have just done that from the start 🤷 

Without this… you could build the types _exactly once_. Repeat builds would confuse TypeScript since the filename prefixes are the same. I.e., it would think we were trying to import our export 🤖

x-template.js
```diff
@@ -0,0 +1,1062 @@
...
+
+  static #interfaceRemoved(name) {
+    return () => {
+      throw new Error(`Removed "${name}" from default templating engine interface. Import and plug-in "lit-html" as your element's templating engine if you want this functionality.`);
+    };
+  }
+}
+
+// Long-term interface.
+export const render = TemplateEngine.interface.render.bind(TemplateEngine);
```
theengineear, line 1040: Rather than export the class (which would lend itself to potential extension / misuse)… we export the exact functions in the interface (many of which are soon to be deleted!).

theengineear:
FYI @klebba — Just working through some of the decisions we’ve been chatting about. In this case, we agreed that splitting this into two files is probably the right move. I went ahead and bumped some dev dependencies as well.

PR #228, opened 2024-12-05
Improve our performance demo.

theengineear:
There is still a fair amount of work to be done before we can be super confident in the results here, but this feels like a good step forward:

* Tests all run in an iframe.
* Progress bar is printed to improve feedback.
* Long text moved to “PERFORMANCE.md”.


theengineear:
@klebba — Just pulling these changes in from some future work I’ve been messing around with in the margins. Mostly just trying to reduce impending git conflicts 😅

PR #231, opened 2024-12-05
Introduce “UnforgivingHtml” parser.

theengineear:
Goals of the parser:
* Tighten control over things like double-quotes & closing tags.
* Improve error messaging for malformed markup.
* Improve performance.

TODOs / Questions:
- [x] Squash commits. The `x-parser.js` commit is left on its own for a moment to figure out if we like that factoring or not.
- [x] … if we keep the above refactoring, update the changelog and the commit message to reflect this.
- [x] Extensive integration testing…
- [x] ~Do we need to strictly reject `CDATA` in `math` and `svg` namespaces? https://w3c.github.io/html-reference/syntax.html#cdata-sections / https://developer.mozilla.org/en-US/docs/Web/API/CDATASection. Note, it appears that when attempting to add CDATA to _html_… it is converted to an html comment. Otherwise, in svg and math — it will work as expected.~ Pass, we can introduce new “forbidden” parsing states to throw an error here. It was already failing due to the fact that it wouldn’t match — but we now add a special error-time lookup to throw a helpful error.
- [x] ~Some of the validations around tag name, attribute name, and property name may be slow… should we be less strict? Is it the case that the browser would throw on any of these anyhow? Could we just let that happen?~ Punting this problem to later — let’s start strict.
- [x] ~I think HTML is often assumed to be UTF-8, but JS strings (i think) are sometimes treated via UTF-16. That means we can accept UTF-16 characters and inject them as text content. But, should we reject that?~ I would prefer to go as long as possible without concerning character encodings. If / when we can think of a way this could break (i.e., differing character encodings in text / files) — we could always revisit.
- [x] ~Is there a way to test UTF-8 multi-byte sequence interop with multi-by UTF-16 surrogate pair interop. Maybe not a problem, but it could be important to know one way or the other!~ I think if you really blow it, you will get some sort of error. See [isWellFormed](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/isWellFormed) for some context.
- [x] ~Do we need special handling of NUL character — U+0000? It is forbidden in text / character data. What about “permanently undefined Unicode characters”?~ YAGNI.
- [x] Can text nodes be added in svg / math namespaces? [yes](https://html.spec.whatwg.org/multipage/syntax.html#foreign-elements)
- [x] ~Should we use setAttributeNS when in a namespace?~ I believe this would only be required if there was a _prefix_ associated with an attribute name like `spec:align`… since we would reject that attribute name anyhow, I think we already can’t get into this situation. Going to not bother with this one.
- [x] ~Do we need special consideration around optional newline for pre / textarea — https://html.spec.whatwg.org/multipage/syntax.html#element-restrictions. Indeed, if you set like `document.body.innerHTML = '<pre>\nhi</pre>` — that initial newline will _not_ appear. Yuck.~ Yes. Fixed. (still “yuck” though!)
- [x] ~Are there any more maps we ought to turn into more-performant switches?~ This is a “later” problem.
- [x] ~Seems like there may be some conventions to use camelCase element names in SVG. Are we okay forcing them to be lowercase?~ Might as well start strict.


x-template.js
```diff
@@ -1,4 +1,5 @@
 /** Forgiving HTML parser which leverages innerHTML. */
+/*
```
klebba, line 2: TODO: re-home `Forgiving` into demo or something for posterity (e.g. not shipped as a supported lib)

x-template.js
```diff
@@ -269,6 +270,1104 @@ class Forgiving {
...
+
+  // Special keys to hang internal state off of nodes we create for performance.
+  static #localName = Symbol();
+  static #parentNode = Symbol();
+  static #namespace = Symbol();
+
+  // Delimiter we add to improve debugging — "${…}".
+  static #delimiter = '${\u2026}';
+
+  // Self-closing html elements.
```
klebba, line 324: TODO consider linking relevant spec stuff contextually vs in top line header?

x-template.js
```diff
@@ -269,6 +270,1104 @@ class Forgiving {
...
+  static #voidHtmlElements = new Set([
+    'area', 'base', 'br', 'col', 'command', 'embed', 'hr', 'img', 'input',
+    'keygen', 'link', 'meta', 'source', 'track', 'wbr',
+  ]);
+
+  // Elements we allow.
+  static #allowedHtmlElements = new Set([
```
klebba, line 331: TODO consider enumerating `disallowedHtmlElements` ?

x-template.js
```diff
@@ -269,6 +270,1104 @@ class Forgiving {
...
+  // by a space or newline. The closing tag must not contain spaces or newlines.
+  // - ok: <h6>, <my-element-1>
+  // - not ok: <-div>, <1-my-element>
+  static #openTagStart = /<(?![0-9-])[a-z0-9-]+(?<!-)(?=[\s\n>])/y;
+  static #closeTag =   /<\/(?![0-9-])[a-z0-9-]+(?<!-)>/y;
+  static #openTagSpace =   /[\s\n]+/y;
+  static #openTagEnd = /(?<![\s\n])>/y;
+
+  // Attribute names can contain the characters [a-z], [A-Z], [0-9], and a
+  // hyphen. But, they cannot begin with numbers, capital letters, or hyphens —
```
klebba, line 561: question:
>they cannot begin with numbers, capital letters, or hyphens
etc — according to us or w3c?

klebba: nit: please note the disparity between the w3c spec standard and our own choices

x-template.js
```diff
@@ -269,6 +270,1104 @@ class Forgiving {
...
+
+  //////////////////////////////////////////////////////////////////////////////
+  // Special Tag Patterns //////////////////////////////////////////////////////
+  //////////////////////////////////////////////////////////////////////////////
+
+  // The “textarea” and “style” tags are special in that they contain so-called
+  // “replaceable” and “non-replaceable” character data. In either case, we
+  // must begin by treating all characters between the opening and closing tags
+  // as their content. Note that we allow the “.” to match across newlines.
+  static #throughStyle =    /.*?<\/style>/ys;
```
klebba, line 593: nit: note that `style` will be deprecated

x-template.js
```diff
@@ -269,6 +270,1104 @@ class Forgiving {
...
+  static #throughTextarea = /.*?<\/textarea>/ys;
+
+  //////////////////////////////////////////////////////////////////////////////
+  // JS-y Unicode Patterns /////////////////////////////////////////////////////
+  //////////////////////////////////////////////////////////////////////////////
+
+  // This style of entering unicode is a JS-ism. We want developers to write
+  // HTML here, not JS. You can of course _interpolate_ whatever you want.
+  static #rawUnicodeOrHexadecimal = /.*[^\\]\\(?:x|u)/ys;
```
klebba, line 602: nit: add an example comment

```
\u2022 // valid javascript, invalid css, nonsense in html
```

x-template.js
```diff
@@ -269,6 +270,1104 @@ class Forgiving {
...
+  //
+  // So-called “replaceable” character data (e.g., most content) can contain
+  // character references (i.e., html entities) which need to be decoded. We
+  // match such entities broadly and then require specific formatting of the
+  // entity thereafter (e.g., alpha hex digits must be upper-case). Such
+  // references can be “named”, “hexadecimal” code points or “decimal” code
+  // points. And, for completeness, large code points can result in multiple
+  // characters as replacement text.
+  static #entity =           /&.*?;/ys;
+  static #hexEntity =        /^&#x[0-9A-F]+;$/;
```
klebba, line 621: e.g. 
```
\0a00 vs \0A00
```

x-template.js
```diff
@@ -269,6 +270,1104 @@ class Forgiving {
...
+  static #decEntity =        /^&#[0-9]+;$/;
```
klebba, line 622: e.g. 

Sign | Name code | Decimal code | Hex code | Description
-- | -- | -- | -- | --
♥ | &hearts; | &#9829; | &#x2665; | heart symbol

x-template.js
```diff
@@ -269,6 +270,1104 @@ class Forgiving {
...
+            `Seems like you have a malformed decimal character reference`,
+            `(html entity). You will need to fix the reference`,
+            `"${encoded}".`,
+          ].join(' '), { cause });
+        }
+      } else {
+        // Because named character references can be written as either a decimal
+        //  or hexadecimal code points — we only support a small subset of named
+        //  entities within this parser.
+        switch (encoded) {
```
klebba, line 997: The reason this might become frustrating is that a developer must remember *why* `&mdash;` works but `&hearts;` does not. I think I will find this surprising; having the engine THROW and EXPLAIN is potentially a good middle ground (for now &hellip...)

klebba, line 997: question: can we enumerate all of the named html entities here?
answer: yes: https://html.spec.whatwg.org/entities.json

Does this list change often?

klebba: Alternatively we could perhaps attempt to let the browser expand html entities using `innerHTML` carefully and selectively

x-template.js
theengineear: Yep already planned — it’s in a comment here so I don’t need to keep resolving difficult git conflicts since we keep pulling functionality and tests forward from this change set and then rebasing this branch back on `main`.

x-template.js
theengineear: I went through and added much more additional commentary. I think this file is more commentary than code 😺

x-template.js
theengineear: I changed to be like:

```js
static #htmlElements = new Set(/* … */)
static #deniedHtmlElements = new Set(/* … */)
static #allowedHtmlElements = Unforgiving.#htmlElements.difference(Unforgiving.#deniedHtmlElements)
```

x-template.js
theengineear: I did my best to capture this detail — I’m sure I’ve not done a perfect job yet, but we may need to simply iterate here…

x-template.js
theengineear: Added.

x-template.js
theengineear: Removed.

x-template.js
theengineear: Removed.

x-template.js
theengineear: Changed, as discussed, this makes us more forgiving for invalid entities, but we decided that was ok.

theengineear: Nope, going a different direction.

x-template.js
theengineear: This is kinda obviated IMO. Not going to worry about it for now.

x-template.js
```diff
@@ -1,271 +1,986 @@
...
+        throw new Error(`[${errorMessagesKey}] ${errorMessage}\n${substringMessage}`);
       }
-      htmlStrings[iii] = string;
+      const encoded = content.slice(contentIndex, Unforgiving.#entity.lastIndex);
+      Unforgiving.#htmlEntityContainer.innerHTML = encoded;
+      const decoded = Unforgiving.#htmlEntityContainer.content.textContent;
+      content = content.replace(encoded, decoded);
+      Unforgiving.#htmlEntityStart.lastIndex = contentIndex + decoded.length;
     }
-    // Again, there might be a quote we need to slice off here still.
```
theengineear, line 127: As mentioned — removing that `Forgiving` comment has made the git diff super wonky here. Let me know if you want that back in 👌

theengineear:
OK @klebba — Certainly, there are still bugs remaining in here, but I believe this should be a reasonable first-pass at the new “unforgiving” parser. I’m going to flip my attention to integration testing this in code we’ve authored to start confirming parsing behavior.

theengineear:
@klebba — going to get this merged in and published _as a release candidate_ so I can do further testing and try to work through #244. Definitely happy to continue to iterate on this stuff though!

klebba:
Great work!

PR #232, opened 2024-12-05
Synchronously parse html / svg.

theengineear:
Previously, the parsing only happened after calling “render”. The issue with this is that you lose the original line numbers on errors when something goes wrong. By forcing the parse to happen immediately, you can find the actual invocation of “html” within which the issue lies.

The only potential downside to this approach would be a performance hit within our “update” path. This is because we would check whether the “strings” from our tagged template function have been analyzed or not. Checking for their existence in our WeakMap takes some non-zero amount of time.

We’re going to go ahead and _try_ this approach and weight the costs against the benefits — it’s easy to revert.


test/test-template-engine.js
```diff
@@ -1094,80 +1094,70 @@ describe('value issues', () => {
...
-    const callback = () => render(container, html`<textarea id="target"><b>please ${'foo'} no</b></textarea>`);
+    const callback = () => html`<textarea id="target"><b>please ${'foo'} no</b></textarea>`;
     const expectedMessage = 'Only basic interpolation of "textarea" tags is allowed.';
     assertThrows(callback, expectedMessage);
   });
 
   it('throws when attempting non-trivial interpolation of a title tag', () => {
-    const container = document.createElement('div');
-    const callback = () => render(container, html`<title>please ${'foo'} no</title>`);
+    const callback = () => html`<title>please ${'foo'} no</title>`;
```
theengineear, line 1130: Note that we no longer need to `render` to see an error. This separation of concerns feels appropriate… but TBD on whether this would ever be a performance bottleneck.

x-template.js
```diff
@@ -875,11 +868,22 @@ class TemplateEngine {
...
+      const fragment = TemplateEngine.#createFragment(language, strings);
+      const lookups = TemplateEngine.#findLookups(fragment);
+      analysis.language = language;
+      analysis.fragment = fragment;
+      analysis.lookups = lookups;
+      analysis.done = true;
+    }
+    // This is a leaking implementation detail, but fixing the leak comes at
+    //  a non-negligible performance cost.
+    return { [TemplateEngine.#ANALYSIS]: analysis, strings, values };
```
theengineear, line 882: Technically, you could `.getOwnPropertySymbols()` to find the non-enumerable “hidden” symbols like the one we use here to gain access to our internal `analysis`. That feels like a bit of a stretch to be concerned about though… practically-speaking.

x-template.js
```diff
@@ -875,11 +868,22 @@ class TemplateEngine {
   }
 
   static #createRawResult(language, strings, values) {
-    return { [TemplateEngine.#LANGUAGE]: language, strings, values };
+    const analysis = TemplateEngine.#setIfMissing(TemplateEngine.#stringsToAnalysis, strings, () => ({}));
```
theengineear, line 871: This is the part that is a potential bottleneck. I’m not seeing any immediate slowdowns in performance — BUT, the performance test really isn’t testing the extremes of this. It’s essentially testing the performance of getting a value from a map with a size of _one_ (since we’re updating with the _same_ strings — that’s what an update is…). It’s possible that when we have a very _large_ number of templates, this could become less performant. We will have _at least_ as many items in this map as there are components in the page (assuming that a component invokes `html` only once — which is typically an under-estimate).

PR #234, opened 2024-12-07
Factor out Forgiving parser from template engine.

theengineear:
We want to experiment with an Unforgiving parser which our template engine can leverage. To enable better side-by-side comparison, we can factor out the current parser from our template engine and give it the same `parse` interface.


x-template.js
```diff
@@ -1,31 +1,290 @@
-/** Internal implementation details for template engine. */
-class TemplateEngine {
+/** Forgiving HTML parser which leverages innerHTML. */
+class Forgiving {
```
theengineear, line 2: Just moving some things around to form a new conceptual boundary between our “parser” and our “dom manager”.

x-template.js
```diff
@@ -870,9 +966,15 @@ class TemplateEngine {
...
-      analysis.language = language;
+      const lookups = {};
+      const onBoolean = TemplateEngine.#storeKeyLookup.bind(null, lookups, TemplateEngine.#BOOLEAN);
+      const onDefined = TemplateEngine.#storeKeyLookup.bind(null, lookups, TemplateEngine.#DEFINED);
+      const onAttribute = TemplateEngine.#storeKeyLookup.bind(null, lookups, TemplateEngine.#ATTRIBUTE);
+      const onProperty =  TemplateEngine.#storeKeyLookup.bind(null, lookups, TemplateEngine.#PROPERTY);
+      const onContent = TemplateEngine.#storeContentLookup.bind(null, lookups);
+      const onText = TemplateEngine.#storeTextLookup.bind(null, lookups);
+      const forgivingLanguage = language === TemplateEngine.#SVG ? Forgiving.svg : Forgiving.html;
+      const fragment = Forgiving.parse(strings, onBoolean, onDefined, onAttribute, onProperty, onContent, onText, forgivingLanguage);
```
theengineear, line 977: This is isn’t an interface we will ever make public. But — the idea is that it’s more like a visitor pattern. the parser just tells the integrator when something “interesting” happens and where it happened. This was one of the untangling changes we made for `Unforgiving` anyway, just back-porting it, really.

theengineear:
FYI @klebba — Thanks for the motivation to improve our ability to do side-by-side comparison here in the future. Now that there is a one-line coupling between `TemplateEngine` and `Forgiving`, it makes it easier in the future to move `Forgiving` into like `/test/forgiving.js` which we can manually hook up as a comparison.

PR #235, opened 2024-12-16
Updates demo styles and improves error messages.

theengineear:
Changes:
* Uses `static get styles()` for demo / test elements. †
* Uses brackets in error messaging — e.g., “<style>” vs “"style"”. ††

†  Inlining style tags will throw a warning in future releases. And, the
   suggested usage is to leverage the static `styles` getter.
†† Rather than use quotes when messaging about tag names, it provides
   developers better context to see messaging that resembles open tags
   so that it’s clear what the error message is all about.

### DIFF
--- a/demo/chess/index.js
+++ b/demo/chess/index.js
@@ -1,6 +1,67 @@
 import XElement from '../../x-element.js';
 
 class ChessPieceElement extends XElement {
+  static get styles() {
+    const styleSheet = new CSSStyleSheet();
+    styleSheet.replaceSync(`\
+      :host {
+        display: block;
+        width: var(--hello-size, 8rem);
+        height: var(--hello-size, 8rem);
+        background-color: cyan;
+        border-radius: 50%;
+        margin: 0.25rem;
+        box-sizing: border-box;
+        transition-duration: 250ms;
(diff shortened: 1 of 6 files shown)


demo/chess/index.js
```diff
@@ -17,65 +78,7 @@ class ChessPieceElement extends XElement {
 
   static template(html) {
     return ({ rank }) => {
-      return html`
-        <style>
```
theengineear, line 21: The bulk of the changes here are stop using inline `<style>` tags — this will throw a deprecation warning in future `1.x` releases (and be strictly forbidden in `2.x`).

x-template.js
```diff
@@ -188,14 +188,14 @@ class Forgiving {
         (localName === 'style' || localName === 'script') &&
         node.textContent.includes(Forgiving.#CONTENT_MARKER)
       ) {
-        throw new Error(`Interpolation of "${localName}" tags is not allowed.`);
+        throw new Error(`Interpolation of <${localName}> tags is not allowed.`);
```
theengineear, line 191: This is the only functional change (and it’s minor). Just changes `"…"` characters to `<…>`.

PR #246, opened 2025-02-12
Allow template interpolation escapes (i.e., “\$”).

theengineear:
We need to allow _some_ JS-y escapes that cannot otherwise be accomplished through standard HTML syntax. We had already allowed the backtick and backslash characters — just need the dollar sign as well.


test/test-parser.js
```diff
@@ -167,6 +167,7 @@ describe('JS-y escapes', () => {
       The \\u2026, \\\\u2026 character is an ellipsis.
       The \\x8230, \\\\x8230 character is also an ellipsis.
       The \\0, \\\\0 character is the nul character.
+      The \${interpolation} character is okay to escape.
```
theengineear, line 170: Gotta be able to print a literal dollar sign.

Issue #120, opened 2022-11-04
Discussion: Enable support for event bindings in Template Engine.


theengineear:
Authors can leverage the `listeners` block or `listen` / `unlisten` methods in `connectedCallback` / `disconnectedCallback` to manage event handling.

**Why we might add it:**

* It may be expected behavior at this point (as it exists in other templating engines)

**Why we might not add it:**

* It’s redundant — you can already achieve this behavior.
* It adds complexity / size to the base element.
* It adds an additional runtime check — this may or may not be a performance drawback.
* In many years of leveraging other templating engines — we've not once used this functionality.

### Proposal

Users would be able to declare something like the following to add an event handler:

```js
static template(html) {
  return ({ onClick }) => {
    return html`<button type="button" @click="${onClick}">Click Me</button>`;
  };
}
```

theengineear:
I'm opposed to adding this functionality, but I wanted to open this issue up since this behavior is common in other templating engines.

chen-ye:
I'll chime in here to say that one thing that template event binding allows for is the ability to bind to dynamically instantiated elements--such as sub-elements in a `repeat/map`.

For the most part, the same thing can logically be achieved using event delegation or custom events that are listened to at the host level, but there are some instances where directly binding is substantially more ergonomic.

For example, consider the following:

```
<table id="table">
  ${rows.map((row) => html`
    <tr>
        <th id=${row.id} @click=${setActiveRow}>
            <label>${row.label}<label><x-indicator enabled=${row.active}></x-indicator>
        </th>...
    </tr>`}
</table>
```

In order to do something logically similar using listen/unlisten, you could delegate the `click` handler to `#table`--however, because the row headers contain sub elements, the handler would need to crawl the `evt.composedPath()` in order to get to the `th`--which is both less efficient, and more brittle.

Alternatively, you could create a custom element for the table row header cell, and rely on css in order to make it behave the same as a native `<th>` element--however that's a substantial amount of boilerplate to just enable individual table row headers to be clicked.

In addition, I'll add that I personally prefer template event binding because it's one fewer opportunity for a mapping error--it's always possible to typo a `listen` call or overlook it when refactoring, but it's impossible to disassociate a template event binding, given that it is on the bound element markup itself.

theengineear:
[...]

PR #252, opened 2025-02-16
Remove `svg` from default template engine.

### LINKED ISSUE #236: Remove support for <svg> elements and “svg” tagged template function in default template engine.
One of the goals of the default template engine is to remain simple. It _must_ enable rendering complex html structures, but the spec for so-called “foreign elements” (e.g., svg and math) diverges quite a bit. For example, attributes are case-insensitive in html, but case-sensitive in svg and math.

The three choices we have are…
1. Continue to support `<svg>` interpolations and the `svg` tagged template function.
2. Support it minimally in the future (e.g., perhaps static, with no interpolations).
3. Remove support (e.g., require that developers inject such things via a document fragment).

Supporting (1) above is quite nuanced as the spec (and conventions) differ significantly. Rather than trying to make _everything_ work — it may be a better choice to simply restrict this option and provide documentation around workarounds.

For example, one of the main uses for svg is static iconography. This can easily be done by setting `template.setHTMLUnsafe('<svg>…</svg>')` and then passing `template.content` (a document fragment) to the template engine as an interpolated value.

theengineear:
The spec and conventions for `svg` differ a lot from `html`. A faster / more-maintainable parser can be built if we stop supporting this.

Closes #236


theengineear:
@klebba — I figured you’d ask about this, so I opened up a PR 🤙

test/test-parser.js
```diff
@@ -560,7 +559,7 @@ describe('errors', () => {
 
   it('throws if you give a custom element tag name with capital letters', () => {
     const callback = () => html`<my-Element></my-Element>`;
-    const expectedMessage = '[#193]';
+    const expectedMessage = '[#120]';
```
theengineear, line 562: Note, the `#120`-range errors are _normal_ — the `#190`-range errors are reserved for deprecations / transient issues that _should_ get delete and go away. I.e., this is a positive change.

x-parser.js
```diff
@@ -208,10 +118,8 @@ export class XParser {
   // Examples:
   //  - ok: <h6>, <my-element-1>
   //  - not ok: <-div>, <1-my-element>
-  // TODO: #236: Restrict tags to lowercase once <svg> support is removed.
-  // Svg tag names follow the above rules, but may have capital letters.
-  #openTagStart = /<(?![0-9A-Z-])[a-zA-Z0-9-]+(?<!-)(?=[\s\n>])/y;
-  #closeTag =   /<\/(?![0-9A-Z-])[a-zA-Z0-9-]+(?<!-)>/y;
+  #openTagStart = /<(?![0-9-])[a-z0-9-]+(?<!-)(?=[\s\n>])/y;
```
theengineear, line 121: This is the reason we get into a different error range. We simply don’t look for _any_ upper case letters in tag names / attribute names anymore.

theengineear:
Casey, gonna keep things moving here and get this merged in. We can revisit this one later as needed.

klebba:
Thanks for taking this up — if we really miss interpolated `svg` support we could always revisit in the future

PR #255, opened 2025-02-18
Prevent disconnects on simple re-mappings.

### LINKED ISSUE #253: Neither `connectedCallback` nor `disconnectedCallback` should be called when re-mapping causes prefix or suffix removal.
Conceptually, if you render `[foo, bar]` and then re-render just `[bar]`, only the `foo` element ought to be disconnected. However, because low-level DOM apis will cause a `disconnect` followed by a `connect` — failure to _carefully_ consider whether re-shuffling or simple truncation is required will cause unexpected lifecycle callbacks.

It’s also unexpected for _shuffling_ to cause these lifecycle callbacks, but we need to wait for better low-level apis. This ticket only covers the _truncation_ (prefix/suffix removal) situations.

theengineear:
Conceptually, if you render `[foo, bar]` and then re-render with just `[bar]` — you don’t expect that the `foo` element will be disconnected and reconnected. However, if we naively first _move_ `bar` to the first slot and then _delete_ `foo`… that‘s exactly what will happen.

Instead, we must first _delete_ `foo` and then _check_ to see if we still need to _move_ `bar`.

Note that if a shuffle is _actually_ necessary, we cannot prevent a disconnect / reconnect from occurring with the current DOM APIs.

Closes #253

### DIFF
--- a/test/test-template-engine.js
+++ b/test/test-template-engine.js
@@ -505,6 +505,46 @@ describe('html rendering', () => {
     assert(!!container2.querySelector('#target'));
   });
 
+  it('causes single connect / disconnect for re-rendered elements', () => {
+    let connects = 0;
+    let disconnects = 0;
+    class TestConnectDisconnect extends HTMLElement {
+      connectedCallback() { connects++; }
+      disconnectedCallback() { disconnects++; }
+    }
+    customElements.define('test-connect-disconnect', TestConnectDisconnect);
+
+    const getTemplate = ({ text }) => {
+      return html`
+        <div id="target">
+          <test-connect-disconnect>${text}</test-connect-disconnect>
+        </div>
+      `;
+    };
+    const container = document.createElement('div');
+    document.body.append(container);
+
+    assert(connects === 0);
+    assert(disconnects === 0);
+
+    render(container, getTemplate({ text: 'foo' }));
+
+    assert(connects === 1);
+    assert(disconnects === 0);
+
+    render(container, getTemplate({ text: 'bar' }));
+
+    assert(connects === 1);
+    assert(disconnects === 0);
+
+    render(container, null);
+
+    assert(connects === 1);
+    assert(disconnects === 1);
+  
+    container.remove();
+  });
+
   it('native map renders basic template', () => {
     const getTemplate = ({ items }) => {
       return html`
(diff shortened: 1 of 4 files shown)


test/test-template-engine.js
```diff
@@ -505,6 +505,46 @@ describe('html rendering', () => {
     assert(!!container2.querySelector('#target'));
   });
 
+  it('causes single connect / disconnect for re-rendered elements', () => {
```
theengineear, line 508: Note, this previously passed – but I wanted to add this for completeness.

test/test-template-engine.js
```diff
@@ -660,6 +700,127 @@ describe('html rendering', () => {
     assert(!!container.querySelector('#foo'));
     assert(container.querySelector('#target').children[0] !== foo);
   });
+
+  it('native map does not cause disconnectedCallback on prefix removal', () => {
```
theengineear, line 704: This failed before — this is the key thing we want to fix.

test/test-template-engine.js
```diff
@@ -660,6 +700,127 @@ describe('html rendering', () => {
...
+    assert(disconnects === 1);
+
+    render(container, getTemplate({ items: [] }));
+    assert(connects === 2);
+    assert(disconnects === 2);
+
+    container.remove();
+  });
+
+  it('native map does not cause disconnectedCallback on suffix removal', () => {
```
theengineear, line 744: This also failed before — this `suffix` test is highly related to the `prefix` test… but again, it feels correct to include for completeness.

test/test-template-engine.js
```diff
@@ -660,6 +700,127 @@ describe('html rendering', () => {
...
+
+    render(container, getTemplate({ items: [] }));
+    assert(connects === 2);
+    assert(disconnects === 2);
+
+    container.remove();
+  });
+
+  // TODO: #254: See https://chromestatus.com/feature/5135990159835136.
+  it.todo('native map does not cause disconnectedCallback on list shuffle', () => {
```
theengineear, line 785: Finally, the `shuffle` test is forthcoming. The `moveBefore` API should solve this if / when we get it. I figured it’d be cool to just include this as a proper `it.todo` which we _should_ be able to simply run later when we improve the underlying code.

theengineear:
This fixes a bug, so I’m going to go ahead and get this merged in.

test/test-template-engine.js
klebba: nice catch!

test/test-template-engine.js
theengineear: Dunno, I only checked against native APIs (e.g., `insertBefore` which does indeed do this).

theengineear: Once something like `moveBefore` lands — any / all templating engines will simply use that and this ought to be a non-issue.

PR #257, opened 2025-02-18
Update README.md installation steps

klebba:
Installation steps must evolve with the introduction of `x-template.js` and `x-parser.js`

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,35 +20,37 @@ A dead simple starting point for custom elements. It provides the following func
 ## Installation:
 
 Save a local copy to your project:
-```
+```bash
 curl https://raw.githubusercontent.com/Netflix/x-element/main/x-element.js > x-element.js
 curl https://raw.githubusercontent.com/Netflix/x-element/main/x-parser.js > x-parser.js
 curl https://raw.githubusercontent.com/Netflix/x-element/main/x-template.js > x-template.js
 ```
 
 Then import it:
-```
+```js
 import XElement from '/path/to/x-element.js';
 ```
 
-...or directly load it from the web:
-```
+...or load it directly from the web:
+```js
 import XElement from 'https://deno.land/x/element/x-element.js';
 ```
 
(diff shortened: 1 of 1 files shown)


klebba:
@theengineear this begs the question: should we be using bare module specifiers to load inner dependencies?

theengineear:
>should we be using bare module specifiers to load inner dependencies

I don’t think that’s a good idea. If you can reliably use a relative path, I think that‘s the appropriate way to go. A bare-module specifier ought to only be used for a _dependency_, right?

README.md
```diff
@@ -19,20 +19,38 @@ A dead simple starting point for custom elements. It provides the following func
 
 ## Installation:
 
+Save a local copy to your project:
 ```
-# save a local copy
 curl https://raw.githubusercontent.com/Netflix/x-element/main/x-element.js > x-element.js
+curl https://raw.githubusercontent.com/Netflix/x-element/main/x-parser.js > x-parser.js
+curl https://raw.githubusercontent.com/Netflix/x-element/main/x-template.js > x-template.js
```
theengineear, line 26: Nice catch!

README.md
```diff
@@ -19,20 +19,38 @@ A dead simple starting point for custom elements. It provides the following func
...
 ```
 
-or
+Then import it:
+```
```
theengineear, line 30: You can tag this with `js` for a highlight?

README.md
```diff
@@ -19,20 +19,38 @@ A dead simple starting point for custom elements. It provides the following func
...
 import XElement from 'https://deno.land/x/element/x-element.js';
 ```
 
-...or if you're fancy:
+...or use an [importmap](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script/type/importmap):
+```
+<script type="importmap">
+  {
+    "imports": {
+      "@element": "https://deno.land/x/element/x-element.js"
```
theengineear, line 44: I’d recommend we use `@netflix/x-element` as the bare module specifier prefix (and then update the below import to match).

theengineear (approved):
LGTM!

PR #259, opened 2025-02-18
Update package.json engines again

klebba:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set/difference — actually this method depends on Node 22 since 2024


theengineear (approved):
Thanks!

PR #260, opened 2025-02-19
Restrict usage of “style” tags in “x-parser”.

### LINKED ISSUE #237: Remove support for <style> elements in default template engine.
Now that the “Import Attributes” TC39 proposal has been accepted, browsers are allowing developers to import instantiated `CSSStyleSheet` objects from `.css` files directly. This obviates the need to inline a `<style>` tag in a template — instead, you can just move it to a `.css` file.

Since (1) the rules to support `<style>` tags are specialized and (2) it’s preferable to externalize style sheets anyhow — this seems like an obvious choice to restrict usage of `<style>` within the default template engine.

theengineear:
The goal here is to continue to reduce the surface area of this feature to be as minimal as is sensible. In this case, basic browser APIs allow authors to add style sheets to shadow roots (or documents) via the `adoptedStyleSheets` setter. The most modern approach here is to write an _actual_ `.css` file and import that as a CSS Module via the new(ish) “import attributes” syntax.

This is much easier for browsers to optimize and simplifies our parsing code a fair amount.

Closes #237


x-parser.js
```diff
@@ -77,12 +73,11 @@ export class XParser {
     'plaintext', 'rb', 'rtc', 'shadow', 'strike', 'tt', 'xmp',
   ]);
   #deniedHtmlElements = new Set([
-    'html', 'head', 'base', 'link', 'meta', 'title', 'body', 'script',
+    'html', 'head', 'base', 'link', 'meta', 'title', 'style', 'body', 'script',
```
theengineear, line 76: 👋

PR #261, opened 2025-02-19
Remove “unsafeHTML” from “x-template” interface.

### LINKED ISSUE #216: Discussion: should we rework or remove unsafeHTML and unsafeSVG
Opening an issue to cover past discussion on this topic — work has already begun in #213 — essentially:

- `unsafeHTML` and `unsafeSVG` are necessary for developers who prefer to avoid imperative DOM manipulation
- it's not necessary to avoid imperative DOM manipulation
- the convenience of injecting `unsafe` markup into an otherwise safe template is debatable
- in our experience `unsafeSVG` has never been necessary (because `unsafeHTML` can accomplish the same behavior when `<svg></svg>` wraps the content)

Thus if we remove `unsafe` entirely we will (potentially inadvertently) encourage unsafe behavior unknowingly (if `innerHTML` were called `unsafeCrazyNativeFeature` we would perhaps feel differently). At the same time, the project philosophy states:

> Presume adopters are browser experts already (stay out of their way)

Which could be interpreted as "stay out of my way, I know `innerHTML` is unsafe!" -or- "stay out of my way, I want to conveniently bind `unsafe` markup in my declarative template!"

The philosophy also states:

> Implement a minimal set of generalized functionality

... and it could be argued that `unsafe` is not a "minimal" feature. Consider the following:

```
<textarea id="example1">${unsafe(mySampleCode)}</textarea>
```

This does not actually touch on the common concern when injecting unsafe HTML — the browser will not actually execute the code within the textarea element — instead it will display the HTML as a string. This snippet is exactly identical:

```
<textarea id="example2">${mySampleCode}</textarea>
```

A more sinister example is:

```
<div id="example3">${unsafe(mySampleCode)}</div>
```

If we had not used the `unsafe` directive, the template engine would simply print the markup as as string (by assigning `div#example3.textContent = mySampleCode` — but this is not what we want. We actually want to allow the browser to expand the string and generate DOM (even if the string contains `<script>/* malicious bad stuff */</script>` because presumably we are confident that other sanitization has already occurred upstream. (Note: do not make that assumption, it is unsafe! Always weigh the risks of what could go wrong and ensure there are other checks in place where it matters.)

So the conundrum in a nutshell:
- get out of the way of experts who presumably know that they need to do unsafe things
- avoid implementing this thing at all; let the experts do the unsafe thing easily but not declaratively (e.g. facilitate getting a pointer to a DOM injection point where the developer can do something unsafe like `div#example4.innerHTML = '<script>/* malicious bad stuff */</script>'`

Open for discussion...

theengineear:
This isn’t commonly used and developers can simply leverage a browser API like `setHTMLUnsafe` to accomplish the same thing by passing a `DocumentFragment` into the templating engine.

Closes #216

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,20 +16,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   element interfaces, “x-template.js” which is all about DOM management, and
   “x-parser.js” which is all about turning markup into a document fragment.
 
-### Deprecated
-
-- The `<style>` tag is deprecated and will be removed in future versions.
-  Authors should prefer to declare a separate stylesheet in a `.css` file now
-  that “import attributes” are supported in modern browsers (#237).
-
 ### Removed
 
+- The `unsafeHTML`, `nullish`, and `map` updaters are gone (#216).
 - The `<style>` element is now restricted. The spec and conventions for `style`
   differ a lot from `html` and a faster / more-maintainable parser can be built
   if we stop supporting this (#237).
-- The `<svg>` element and the `svg` tagged template function are gone. The spec
-  and conventions for `svg` differ a lot from `html` and a faster /
-  more-maintainable parser can be built if we stop supporting this (#236).
+- The `<svg>` element, the `unsafeSVG` updater, and the `svg` tagged template
+  function are gone. The spec and conventions for `svg` differ a lot from `html`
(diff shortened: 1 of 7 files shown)


theengineear:
Closing in favor of #262 which covers both `unsafeHTML` and `live` in a single change set.

theengineear:
Closing in favor of #262 which covers both `unsafeHTML` and `live` in a single change set.

PR #262, opened 2025-02-19
Remove “unsafeHTML” and “live” from interface.

### LINKED ISSUE #208: Discussion: Should we remove “live” from the default template interface?
**Pros (for removal)**
* Removing it means less surface area / forks.
* It’s potentially error-prone / hard to reason about (see **context** below).
* It might force better architectural practices.

**Cons (for removal)**
* It’s kinda hard to roll your own if you need this sort of thing.

**Options**
* Just delete it from the interface — figure it out, bye (see **rolling your own** below)!
* Add a way to register custom updaters (see [#132](https://github.com/Netflix/x-element/issues/132)).

**Current usage**

_With_ `live`, you could do something like this:

```js
static template(html, { live }) {
  return ({ value }) => {
    return html`<input id="input" .value="${live(value)}">`;
  };
}
```

**Rolling your own**

_Without_ `live`, you would have to do something like this:

```js
static template(html) {
  return ({ value }) => {
    return html`<input id="input">`;
  };
}

// After every render, make sure that current values are bound, even if they
//  haven’t changed from the perspective of the outer element.
render() {
  super.render();
  const input = this.shadowRoot.getElementById('input');
  if (input.value !== this.value) {
    input.value = this.value;
  }
}
```

**Context**

Under-the-hood, the `live` binding works by bypassing the `value !== lastValue` equality check. But, this only occurs _on render_. In other words, if the node-in-question changes and a re-render isn’t called… nothing happens. Even worse… since rendering is itself guarded by change-by-reference detection — there’s no guarantee that a top-level re-render would even make its way back to the `live` binding!

The reason a `live` binding is used is for the “controlled form” paradigm. The goal here is to have a single state store that fully declares the state of the form. But of course, there’s state in the DOM too (_especially_ for form controls) — so we end up with competing sources of truth.

One alternative pattern is to let the form state live in the form (that kinda is how it’s designed in the first place) and avoid reentrance there altogether. But, that can be a tricky pattern as well.

### LINKED ISSUE #216: Discussion: should we rework or remove unsafeHTML and unsafeSVG

theengineear:
These aren’t commonly used and developers can simply leverage browser APIs to accomplish these behaviors via `setHTMLUnsafe` and listening to `change` events, respectively.

Closes #208
Closes #216

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,20 +16,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   element interfaces, “x-template.js” which is all about DOM management, and
   “x-parser.js” which is all about turning markup into a document fragment.
 
-### Deprecated
-
-- The `<style>` tag is deprecated and will be removed in future versions.
-  Authors should prefer to declare a separate stylesheet in a `.css` file now
-  that “import attributes” are supported in modern browsers (#237).
-
 ### Removed
 
+- The `unsafeHTML`, `live`, `nullish`, and `map` updaters are gone (#208, #216).
 - The `<style>` element is now restricted. The spec and conventions for `style`
   differ a lot from `html` and a faster / more-maintainable parser can be built
   if we stop supporting this (#237).
-- The `<svg>` element and the `svg` tagged template function are gone. The spec
-  and conventions for `svg` differ a lot from `html` and a faster /
-  more-maintainable parser can be built if we stop supporting this (#236).
+- The `<svg>` element, the `unsafeSVG` updater, and the `svg` tagged template
+  function are gone. The spec and conventions for `svg` differ a lot from `html`
+  and a faster / more-maintainable parser can be built if we stop supporting
+  this (#236).
 - Support for the `<math>` element is removed from the default template engine.
   This worked before because `innerHTML` was being used under-the-hood. But a
   strict allow-list is now used to accomplish parsing (#238).
--- a/x-template.js
+++ b/x-template.js
@@ -39,11 +39,7 @@ class TemplateEngine {
     html: TemplateEngine.html,
 
     // Deprecated interface.
-    map: TemplateEngine.#interfaceDeprecated('map', TemplateEngine.map),
-    live: TemplateEngine.#interfaceDeprecated('live', TemplateEngine.live),
-    unsafeHTML: TemplateEngine.#interfaceDeprecated('unsafeHTML', TemplateEngine.unsafeHTML),
     ifDefined: TemplateEngine.#interfaceDeprecated('ifDefined', TemplateEngine.ifDefined),
-    nullish: TemplateEngine.#interfaceDeprecated('nullish', TemplateEngine.nullish),
     repeat: TemplateEngine.#interfaceDeprecated('repeat', TemplateEngine.repeat),
   });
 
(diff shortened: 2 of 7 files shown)


CHANGELOG.md
```diff
@@ -16,20 +16,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   element interfaces, “x-template.js” which is all about DOM management, and
   “x-parser.js” which is all about turning markup into a document fragment.
 
-### Deprecated
-
-- The `<style>` tag is deprecated and will be removed in future versions.
```
theengineear, line 21: This is actually just removed — forgot to delete this line item in a prior PR.

x-template.js
```diff
@@ -780,18 +633,8 @@ class TemplateEngine {
     }
   }
 
-  static #throwUpdaterError(updater, binding) {
```
theengineear, line 783: We’re down to _just one_ “updater”, which is `ifDefined`. By `2.x` we will remove this as well an expect that authors will leverage the `??` binding to get the same effect.

theengineear:
FYI @klebba — Just moving things along. This will be the last interface removal for a little bit. But again, we do intend to remove all the deprecated things ahead of the actual `2.x` release.

PR #268, opened 2025-02-20
Remove “ifDefined” content guards.

theengineear:
Developers will no longer get errors thrown related to misused of the “ifDefined” updater. This is not really a case that we need to guard against though as it’s not ever mistakenly done in practice.

Note that the guard _does_ remain in place for now for boolean attributes, defined attributes, and properties (which are common places where mistakes occur). In the future, “ifDefined” will be deleted completely and we’ll be able to do even more cleanup here.

Also, because we removed the “live” in a prior PR, we are also able to move the change-by-reference detection guard up to the top-level “commit” method for improved readability and performance.

### DIFF
--- a/x-template.js
+++ b/x-template.js
@@ -296,49 +296,40 @@ class TemplateEngine {
       const lastUpdate = TemplateEngine.#symbolToUpdate.get(lastValue);
       TemplateEngine.#ifDefined(node, name, update.value, lastUpdate?.value);
     } else {
-      if (value !== lastValue) {
-        node.setAttribute(name, value);
-      }
+      node.setAttribute(name, value);
     }
   }
 
-  static #commitBoolean(node, name, value, lastValue) {
+  static #commitBoolean(node, name, value) {
     const update = TemplateEngine.#symbolToUpdate.get(value);
     if (update) {
       TemplateEngine.#throwIfDefinedError(TemplateEngine.#BOOLEAN);
     } else {
-      if (value !== lastValue) {
-        value ? node.setAttribute(name, '') : node.removeAttribute(name);
-      }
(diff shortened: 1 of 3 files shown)


x-template.js
```diff
@@ -353,149 +344,62 @@ class TemplateEngine {
...
+    if (category !== lastCategory && lastValue !== TemplateEngine.#UNSET) {
       // Reset content under certain conditions. E.g., `map` >> `null`.
       const state = TemplateEngine.#getState(node, TemplateEngine.#STATE);
       const arrayState = TemplateEngine.#getState(startNode, TemplateEngine.#ARRAY_STATE);
       TemplateEngine.#removeBetween(startNode, node);
       TemplateEngine.#clearObject(state);
       TemplateEngine.#clearObject(arrayState);
     }
-    if (introspection?.category === 'update') {
-      TemplateEngine.#throwIfDefinedError(TemplateEngine.#CONTENT);
```
theengineear, line 426: Much of the cleanup here simply comes from removing this error. Now, if you _do_ mistakenly wrap a content / text binding in `ifDefined()` — you’ll just see like `function() {}` printed in the DOM. In practice, no one really makes this mistake though…

Because there are no more content updaters _at all_ (previously, `repeat` was an updater, for example) — the cleanup here is more valuable than the possible developer warning.

x-template.js
```diff
@@ -638,8 +522,6 @@ class TemplateEngine {
   }
 
   static #canReuseDom(preparedResult, rawResult) {
-    // TODO: Is it possible that we might have the same strings from a different
```
theengineear, line 641: Because we no longer support svg, we can delete this TODO. There are no other languages than html and so this check is perfectly sufficient (without qualification).

theengineear:
FYI @klebba

PR #271, opened 2025-03-01
Remove environmental fork from “x-parser” logic.

theengineear:
This change set moves _all_ DOM usage out of “x-parser” and into the “x-template” file. Because “x-parser” is used both in Node and Browser environments — we had previously needed to “mock” browser-only APIs for our related “eslint-plugin-x-element” tool.

_Now_, the parser simply produces a stream of “tokens” as it interprets interpolated html markup which can be consumed equally by the related linter and by the default template engine.

It’s a critical design choice to make sure that the code performing the linting is _byte for byte_ the _exact_ same as the code validating html markup in the runtime — otherwise, we open up the door to unpredictable and difficult-to-reproduce issues.†

Additionally, now that we don’t have forking logic in “x-parser”, we no longer encounter Node / ESLint issues if `XParser` is a purely `static` interface again. Therefore, this change set _also_ migrates us back to a static interface such that integrators should not need to instantiate an instance of the parser.

† “Yes” — A local Node version is not byte-for-byte the same as some
  local browser version, but our “x-parser.js” file will be the exact
  same in both places. That’s as much as we can control.


CHANGELOG.md
```diff
@@ -14,7 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   and the usage of spaces and newlines (#239).
 - There are now three files in this repo — “x-element.js” which is all about
   element interfaces, “x-template.js” which is all about DOM management, and
-  “x-parser.js” which is all about turning markup into a document fragment.
+  “x-parser.js” which is all about interpolated html markup interpretation.
```
theengineear, line 17: This `CHANGELOG.md` diff is an excellent summary of the key change.

test/test-parser.js
```diff
@@ -1,73 +1,75 @@
 import { XParser } from '../x-parser.js';
 import { assert, describe, it } from './x-test.js';
 
-const parser = new XParser();
-
-// To prevent DOM creation or to run in non-browser environments, validation can
-//  be made to work by injecting a mock window object.
-class MockWindow {
```
theengineear, line 8: No more mocking needed for validation purposes 🎉

test/test-parser.js
```diff
@@ -85,115 +87,214 @@ const assertThrows = (callback, expectedMessage, options) => {
...
+    const expectedTokens = [];
+    const tokens = html``;
+    assert(deepEqual(tokens, expectedTokens), stringifyTokens(tokens));
   });
 
   it('single string works', () => {
-    const fragment = html`<div id="target">No interpolation.</div>`;
-    assert(fragment.childNodes.length === 1);
-    assert(fragment.querySelector('#target').textContent === 'No interpolation.');
+    const expectedTokens = [
```
theengineear, line 118: The test suite for the parser is more verbose now, but it’s also very exhaustive. It’s quite clear now that we’re interpreting the interpreted markup as expected.

test/test-template-engine.js
```diff
@@ -53,13 +53,66 @@ describe('html rendering', () => {
...
+  });
+
   it('renders named html entities in text content', () => {
     const container = document.createElement('div');
     render(container, html`<div>${'--'}&#123;&lt;&amp;&gt;&apos;&quot;&#x007D;${'--'}</div>`);
     assert(container.childNodes.length === 1);
     assert(container.childNodes[0].textContent === `--{<&>'"}--`);
   });
 
+  it('renders direct references in replaceable character data', () => {
```
theengineear, line 92: Some interesting DOM-specific test cases are moved from the parser tests to the template engine tests because the _require rendering_ to test for correctness.

For example, we can test that character references are _interpreted_ in the parser, but we must ensure that they are then properly _decoded_ in the template engine (which requires DOM APIs).

x-template.js
```diff
@@ -23,6 +21,12 @@ class TemplateEngine {
   static #STATE = Symbol();
   static #ARRAY_STATE = Symbol();
 
+  // It’s more performant to clone a single fragment, so we keep a reference.
+  static #fragment = new DocumentFragment();
```
theengineear, line 25: Note that the creation of the document fragment root is moved to “x-template.js” now. And, there’s no need to conditionally create it as “x-template” can safely assume that it will always be run in a browser.

test/test-parser.js
```diff
@@ -1080,8 +1381,10 @@ describe('html error formatting', () => {
 });
 
 describe('validate', () => {
+  // For simple validation purposes — a noop “onToken” can be provided.
+  const onToken = () => {};
   // eslint-disable-next-line no-shadow
-  const html = strings => wrapper(validator, strings);
+  const html = strings => XParser.parse(strings, onToken);
```
theengineear, line 1387: It’s worth calling out how _simple_ the logic to create a “validator” becomes here.

x-template.js
```diff
@@ -135,6 +139,134 @@ class TemplateEngine {
...
+        state.encoded = false;
+        state.text = '';
+        break;
+      }
+      case XParser.tokenTypes.boundTextValue:
+        onText(state.path);
+        break;
+      case XParser.tokenTypes.boundContentValue:
+        // @ts-ignore — TypeScript doesn’t get that this is an element.
+        state.element.append(document.createComment(''), document.createComment(''));
```
theengineear, line 245: Note that comment cursor creation is moved to “x-template” as they have nothing to do with the markup interpretation. They are a live reference within instantiated DOM that we use as an optimization.

theengineear:
FYI @klebba — As discussed, just getting some of these larger changes in and we can plan to do a more detailed / line-by-line review ahead of the actual `2.0.0` release.

PR #272, opened 2025-03-01
Fixes tests for GitHub actions.

theengineear:
Previously, the tests weren’t actually being run by GitHub actions for some time now. They would run locally, but the browser would silently fail to launch in GitHub when the action was run.

Additionally, we now “Bail out!” within “test.js” if any exceptions are caught — these are magic words that our TAP parser will recognize and they instruct the parser to immediately consider the test a failure.

### DIFF
--- a/test.js
+++ b/test.js
@@ -5,14 +5,9 @@ import puppeteer from 'puppeteer';
     // Open our browser.
     const browser = await puppeteer.launch({
       timeout: 10000,
-      // opt-in to the new Chrome headless implementation
-      // ref: https://developer.chrome.com/articles/new-headless/
-      headless: 'new',
-      args: [
-        // Disables interactive prompt: Do you want to the application Chromium.app to accept incoming network connections?
-        // ref: https://github.com/puppeteer/puppeteer/issues/4752#issuecomment-586599843
-        '--disable-features=DialMediaRouteProvider',
-      ],
+      // These args fix “Error: Failed to launch the browser process!” when
+      //  run in GitHub. We are only running our own code here, so this is ok.
+      args: ['--no-sandbox', '--disable-setuid-sandbox'],
     });
     const page = await browser.newPage();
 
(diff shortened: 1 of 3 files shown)


test.js
```diff
@@ -5,14 +5,9 @@ import puppeteer from 'puppeteer';
     // Open our browser.
     const browser = await puppeteer.launch({
       timeout: 10000,
-      // opt-in to the new Chrome headless implementation
-      // ref: https://developer.chrome.com/articles/new-headless/
-      headless: 'new',
```
theengineear, line 10: I think this setting is outdated now?

test.js
```diff
@@ -5,14 +5,9 @@ import puppeteer from 'puppeteer';
...
-      args: [
-        // Disables interactive prompt: Do you want to the application Chromium.app to accept incoming network connections?
-        // ref: https://github.com/puppeteer/puppeteer/issues/4752#issuecomment-586599843
-        '--disable-features=DialMediaRouteProvider',
```
theengineear, line 14: I checked the linked issue and it seemed like it was closed.

test.js
```diff
@@ -5,14 +5,9 @@ import puppeteer from 'puppeteer';
...
-      ],
+      // These args fix “Error: Failed to launch the browser process!” when
+      //  run in GitHub. We are only running our own code here, so this is ok.
+      args: ['--no-sandbox', '--disable-setuid-sandbox'],
```
theengineear, line 10: Would love a cross-check on this one. I think this is what we need to do now in GitHub based on our settings?

theengineear:
@klebba — Would you mind taking a look at this?

theengineear:
Back to passing now that underlying tests are _actually_ fixed (versus just not running at all).

theengineear:
Note that in addition to _fixing_ this action, we also ensure that it will _break_ in the right way in the future if the browser fails to launch.

Screenshot 2025-03-05 at 7 47 02 PM

klebba (approved):
LG2M

PR #274, opened 2025-03-01
Use import maps in test pages.

### LINKED ISSUE #266: Use bare module specifiers throughout tests to link to @netflix/x-test
This will be necessary for all the reasons

(relates to #265)

theengineear:
Just cleaning up an old “TODO” on this one.

Closes #266.


test/index.html
```diff
@@ -1,8 +1,13 @@
 <!doctype html>
 <html>
+  <head>
+    <meta charset="utf-8">
+    <script type="importmap">
+      { "imports": { "@netflix/x-test/": "../node_modules/@netflix/x-test/" } }
+    </script>
+  </head>
   <body>
-    <meta charset="UTF-8">
```
theengineear, line 4: No need to shout… similar to our usage of `div` instead of `DIV` or `!doctype` instead of `!DOCTYPE` — we typically prefer lowercase when things are case-insensitive.

test/index.html
```diff
@@ -1,8 +1,13 @@
 <!doctype html>
 <html>
+  <head>
+    <meta charset="utf-8">
+    <script type="importmap">
```
theengineear, line 5: This is the main change — just uses an import map which points to our local node modules.

theengineear:
FYI @klebba — We can fuss over some of the details here later, but I wanted to make the major change to move us over to using import maps.

PR #276, opened 2025-03-01
Replaces “innerHTML” with “setHTMLUnsafe”.

theengineear:
Just leverage the more modern approach in the handful of places that use this in our code (for decoding character references and in tests).

This feature was baselined in 2024.

### DIFF
--- a/x-template.js
+++ b/x-template.js
@@ -142,7 +142,7 @@ class TemplateEngine {
 
   // We only decode things we know to be encoded since it’s non-performant.
   static #decode(encoded) {
-    this.#htmlEntityContainer.innerHTML = encoded;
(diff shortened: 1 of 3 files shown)


theengineear:
Just keeping it classy here in 2025.

x-template.js
```diff
@@ -142,7 +142,7 @@ class TemplateEngine {
 
   // We only decode things we know to be encoded since it’s non-performant.
   static #decode(encoded) {
-    this.#htmlEntityContainer.innerHTML = encoded;
+    this.#htmlEntityContainer.setHTMLUnsafe(encoded);
```
theengineear, line 145: This was the only place _in our production code_ that used this. Simple swap.

PR #277, opened 2025-03-01
Adds a test related to literal property bindings.

### LINKED ISSUE #203: Default template engine should throw for literal property binding.
I think I’ve only ever seen a literal property binding _by mistake_, so I’m tempted to keep this strict. However, I wanted to farm for dissent a little bit. Currently…

**NOT OK**

```js
html`<div .foo="bar"></div>`;
```

**OK**

```js
html`<div foo="bar"></div>`;
```

or…

```js
html`<div foo="${'bar'}"></div>`;
```

^^ But this last one is considered an anti-pattern and should not be used.

theengineear:
The default template engine doesn’t allow binding a _literal_ value to a property via the `.` prefix — however, we didn’t previously check this case in our test suite.

Closes #203.

### DIFF
--- a/test/forgiving.js
+++ b/test/forgiving.js
@@ -1,5 +1,3 @@
-// TODO: The `XParser` interface now requires instantiation.
-
 // This is just kept here as an example alternative to our more “unforgiving”
 //  parsing solution. In particular, it could be interesting to try and keep the
 //  interfaces to both “forgiving” and “unforgiving” as similar as possible to
--- a/test/test-parser.js
+++ b/test/test-parser.js
@@ -1085,6 +1085,12 @@ describe('errors', () => {
     assertThrows(callback, expectedMessage, { startsWith: true });
   });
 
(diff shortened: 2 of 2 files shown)


test/forgiving.js
```diff
@@ -1,5 +1,3 @@
-// TODO: The `XParser` interface now requires instantiation.
```
theengineear, line 1: This is no longer true. Just cleaning up this minor note.

test/test-parser.js
```diff
@@ -1085,6 +1085,12 @@ describe('errors', () => {
     assertThrows(callback, expectedMessage, { startsWith: true });
   });
 
+  it('throws for misuse of a bound property prefix with a literal value', () => {
+    const callback = () => html`<div .literal="property"></div>`;
```
theengineear, line 1089: Because this is typically a bug — we don’t allow such bindings at all.

theengineear:
FYI @klebba — no production code changes here, just adding a missing test case.

PR #278, opened 2025-03-02
Improve error context object.

theengineear:
We return some basic location information when errors occur. This change set makes the naming conventions returned in the error context match the naming seen in the `onToken` callback param naming.


x-parser.js
```diff
@@ -1062,13 +1061,18 @@ export class XParser {
...
       }
 
       XParser.#validateExit(tagName);
     } catch (error) {
-      error[XParser.#errorContextKey] = { stringsIndex, string, stringIndex };
+      // Roughly match the conventions for “onToken”.
+      const index = stringsIndex;
+      const start = stringIndex;
+      const end = nextStringIndex;
+      error[XParser.#errorContextKey] = { index, start, end };
```
theengineear, line 1075: `index`, `start` and `end` are the names we use for `onToken` — seems like a good idea to reuse them.

x-parser.js
```diff
@@ -1062,13 +1061,18 @@ export class XParser {
             }
           }
           stringIndex = nextStringIndex; // Update out pointer from our pattern match.
+          nextStringIndex = null;
```
theengineear, line 1064: Null-out the `nextStringIndex` (i.e., `end`) for the next loop. Now that we report it, we want it to be roughly correct.

x-parser.js
```diff
@@ -983,7 +983,6 @@ export class XParser {
                 XParser.#sendBoundContentTokens(onToken, stringsIndex - 1, string, stringIndex);
               }
               value = XParser.#boundContent;
-              nextStringIndex = value.lastIndex;
```
theengineear, line 986: This was not actually doing anything. Removed.

theengineear:
Super minor change, should only impact the linter.

PR #280, opened 2025-03-04
Restrict all JS-y escapes.

### LINKED ISSUE #270: Properly index based on “raw” (versus “cooked”) strings in “x-parser”.
There are two options here…

1. Don’t allow _any_ escapes, i.e., write a back-tick as `&grave;`, backslash as `&bsol;`, and dollar as `&dollar;`.
2. Manually walk over the `raw` string and manage character collapsing accordingly. E.g., if you see `\\` in the raw, you know that’s `\` in the cooked version — this is potentially quite cumbersome… potentially even slow.

Note that if we strictly enforce that _no_ escapes (i.e., no `\` characters _at all_) — then `raw === cooked` and this problem goes away.

One case I can think of that would break is for a `\` character _in a comment_. Because the related character reference _won’t be evaluated_ in a comment and we _don’t allow comment interpolation_ — you’d have to work pretty hard to make it happen.

Does this case matter? Not really… I think part of the goal here is to restrict functionality if / when it could reduce complexity and improve performance. Not being able to write a comment like `<!-- the file is "C:\Foo\Bar" -->` in your markup will probably be _ok_.

theengineear:
We previously allowed escaping the “\”, “$” and “`” characters to accommodate for usage in template literals. However, this causes two important forks (which has complexity and maintenance overhead):

 1. The “raw” and “cooked” strings cannot be assumed to be equivalent. This is a problem because linters that work with “raw” text in an IDE will need different indexing from outputs in the runtime (where text is “cooked”).
 2. Handling of escapes for certain characters differs from others. This is an issue because it increases our pattern matching logic which decreases performance and adds conceptual overhead to understand that _certain_ escaping was allowed, but _other_ escaping wasn’t.

The lowest-code fix for these conceptual forks is to simply _not allow_ any JS-y character escapes. That forces the “cooked” string to be exactly the same as the “raw” string. It also greatly simplifies our validation logic and gives us a small, but non-negligible perf win.

Closes #270.


test/test-parser.js
```diff
@@ -298,22 +298,22 @@ describe('character references', () => {
...
-      This \\, \\\\ character is a literal backslash.
-      The \\t, \\\\t character is a tab.
-      The \\b, \\\\b character is a backspace.
-      The \\f, \\\\f character is a form feed.
-      The \\v, \\\\v character is a vertical tab.
-      The \\u2026, \\\\u2026 character is an ellipsis.
-      The \\x8230, \\\\x8230 character is also an ellipsis.
-      The \\0, \\\\0 character is the nul character.
-      The \${interpolation} character is okay to escape.
+    html`
```
theengineear, line 304: IMO — writing actual HTML here is A-OK with me. As proof of that… I was even able to use `html` here again (instead of `htmlol`) because my syntax highlighter stopped being confused by  the non-HTML-y escape usage of back-slashes.

test/test-parser.js
```diff
@@ -1146,77 +1148,101 @@ describe('errors', () => {
   });
 
   it('throws for trying to write unicode in a js-y decimal format', () => {
-    const callback1 = () => html`<div>please no\x8230</div>`;
-    const callback2 = () => html`<div>please no\\\x8230</div>`;
+    const callback1 = () => htmlol`<div>please no\x8230</div>`;
```
theengineear, line 1151: I updated the tags here since it prevents my syntax highlighter from trying to interpret this as HTML — it really _isn’t_.

test/test-parser.js
```diff
@@ -1146,77 +1148,101 @@ describe('errors', () => {
...
+    assertThrows(callback1, expectedMessage, { startsWith: true });
+  });
+
+  it('throws for trying to write literal back tick via an escape', () => {
+    const callback1 = () => htmlol`<div>\`</div>`;
+    const expectedMessage = '[#150]';
+    assertThrows(callback1, expectedMessage, { startsWith: true });
+  });
+
+  it('throws for trying to write literal dollar sign via an escape', () => {
```
theengineear, line 1240: Added specific tests for the JS-y escapes we _used_ to allow.

test/test-template-engine.js
```diff
@@ -354,16 +354,29 @@ describe('html rendering', () => {
     assert(container.querySelector('#target').prop === 'bar');
   });
 
+  it('renders “on*” attributes as event handlers', () => {
```
theengineear, line 357: Thought it was worth adding related coverage to `on*` _attributes_ usage (versus just properties). The interface is subtly different (and confusing) in that you more-or-less are binding the function _content_ with the _attribute_ and a function _reference_ with the _property_. And, because properties cannot be literal strings in our DSL, there’s an extra layer of nuance.

To lock this down… I just added an extra test. Why not.

x-parser.js
```diff
@@ -551,15 +531,24 @@ export class XParser {
...
+  //  real HTML. Another reason to reject JS-y unicode is that it won’t be
+  //  interpreted correctly by tooling that expects _html_.
+  // Escapes for the “$”, “\”, and “`” characters are required within a template
+  //  literal — users will need to use “&dollar;”, “&bsol;”, and “&grave;”.
+  // Examples:
+  //  - ok: html`&#8230;`, html`&#x2026;`, html`&mldr;`, html`&hellip;`
+  //  - not ok: html`\nhi\nthere`, html`\x8230`, html`\u2026`, html`\s\t\o\p\ \i\t\.`
   static #validateRawString(rawString) {
-    XParser.#rawJsEscape.lastIndex = 0;
-    if (XParser.#rawJsEscape.test(rawString)) {
```
theengineear, line 559: This is the performance gain. Again, small, but non-negligible. A couple microseconds per interpretation. It’s also much easier to quickly grok what this is doing with `rawString.indexOf('\\')`.

x-parser.js
```diff
@@ -168,26 +168,6 @@ export class XParser {
...
-  //  real HTML. Another reason to reject JS-y unicode is that it won’t be
-  //  interpreted correctly by tooling that expects _html_.
-  // The only escapes we expect to see are for the “$”, “\”, and “`” characters,
-  //  which you _must_ use if you need those literal characters.
-  // The simplest way to check this is to ensure that back slashes always come
-  //  in pairs of two or a single back slash preceding a back tick.
-  // Examples:
-  //  - ok: html`&#8230;`, html`&#x2026;`, html`&mldr;`, html`&hellip;`, html`\\n`
-  //  - not ok: html`\nhi\nthere`, html`\x8230`, html`\u2026`, html`\s\t\o\p\ \i\t\.`
-  static #rawJsEscape = /.*(?<!\\)(?:\\{2})*\\(?![$\\`])/ys;
```
theengineear, line 189: Again, this is the complexity that we’re ditching here.

theengineear:
FYI @klebba — Just circling back on a conversation from earlier.

Issue #254, opened 2025-02-18
Neither connectedCallback nor disconnectedCallback should be called when re-mapping causes a shuffle.


theengineear:
Conceptually, if you render `[foo, bar]` and then re-render `[bar, foo]`, neither of those elements should be disconnected at all. However, because low-level DOM apis will cause a `disconnect` followed by a `connect` — there’s no great way to perform this shuffle without causing _at least_ one disconnection.

A new API is being proposed called `moveBefore` — https://github.com/whatwg/dom/pull/1307. And, Chrome has begun implementing it — https://chromestatus.com/feature/5135990159835136. This should directly address the shuffle-disconnect issue.

theengineear:
Note — we should look into adding a stub for `connectedMoveCallback` in the `XElement` base class once this all lands.

theengineear:
Firefox now supports `moveBefore` as well — https://www.firefox.com/en-US/firefox/144.0/releasenotes/

theengineear:
Looks like there is a person assigned to that bug ticket now and things are being updated 🤞

PR #282, opened 2025-03-05
Prepare codebase for new “moveBefore” DOM API.

theengineear:
Things seem to be _moving_ along pretty quickly for this new API. The motivation for this (versus `insertBefore`) is that DOM state is not reset when you _move_ DOM nodes around within the same connected tree.

Importantly, that means a `disconnectedCallback` followed immediately by a `connectedCallback` will no longer happen. Those two lifecycle events typically do some heavy lifting and we want to avoid them for cases like nodes simply being moved around within the same list or something.

Note that this change set _prepares_ us to leverage this new API, but it doesn’t yet commit to it as support is only currently in Chrome 133. We wouldn’t normally get ahead of ourselves like this — but there are a lot of positive signals from other browsers, so it seems like this will go pretty quickly.

Final note — I was able to confirm that shuffling a keyed array won’t cause disconnect / connect callbacks in the future!


theengineear:
Some places to watch:

* WHATWG spec: https://github.com/whatwg/dom/pull/1307
* Chrome Platform Status: https://chromestatus.com/feature/5135990159835136

x-element.js
```diff
@@ -172,6 +172,12 @@ export default class XElement extends HTMLElement {
     XElement.#connectHost(this);
   }
 
+  // TODO: #254: Uncomment once we leverage “moveBefore”.
+  // /**
+  //  * Extends HTMLElement.prototype.connectedMoveCallback.
+  //  */
+  // connectedMoveCallback() {}
```
theengineear, line 179: We’ll eventually want to add this to the base element for completeness / self-documentation, similar to other lifecycle methods.

x-template.js
```diff
@@ -526,6 +604,25 @@ class TemplateEngine {
...
+    const lastCategory = TemplateEngine.#getCategory(lastValue);
+    if (category !== lastCategory && lastValue !== TemplateEngine.#UNSET) {
+      // Reset content under certain conditions. E.g., `map` >> `null`.
+      const state = TemplateEngine.#getState(node, TemplateEngine.#STATE);
+      const arrayState = TemplateEngine.#getState(startNode, TemplateEngine.#ARRAY_STATE);
+      TemplateEngine.#removeBetween(startNode, node);
+      TemplateEngine.#clearObject(state);
+      TemplateEngine.#clearObject(arrayState);
+    }
+    switch (category) {
```
theengineear, line 620: This is the crux of the change — we support _five_ value types which can be bound into a content interpolation. Previously, the `map` and `array` where sort of _mashed together_ into a single handler. However, the internal forks in that logic ended up being a bad factoring.

For example, now that `array` is handled on its own — it’s much easier to read (since it’s by far the simpler of the two cases).

x-template.js
```diff
@@ -477,40 +370,225 @@ class TemplateEngine {
...
       }
-      node.parentNode.insertBefore(value, node);
-    } else {
+      lastItem = item;
+    }
+  }
+
+  // TODO: #254: Future state where the “moveBefore” API is better-supported.
+  // // Loops over given array of “entries” to manage an array of nodes.
+  // static #commitContentArrayEntries(node, startNode, entries) {
```
theengineear, line 529: Eventually (perhaps soon?), we’ll just uncomment this and delete the above handler. This pattern of encoding _what’s next_ has been quite helpful (assuming these comments don’t linger indefinitely).

test/test-template-engine.js
```diff
@@ -867,9 +881,11 @@ describe('html rendering', () => {
   // TODO: #254: See https://chromestatus.com/feature/5135990159835136.
   it.todo('native map does not cause disconnectedCallback on list shuffle', () => {
```
theengineear, line 882: Again — I did test this all out and _indeed_ it does do the thing. One curiosity is that without the `connectedMoveCallback` registered, the behavior is _different_. My guess is that this is not the final behavior, but I will probably ask around to see if I can get more information there.

x-template.js
theengineear: My experience with trying to do that has just been stale PRs. I’m going to leave this as-is for now, but the goal will be to take up all these commented-out code blocks _ahead_ of the `2.x` release.

So, if it turns out we really cannot take this on ahead of `2.x` I’ll likely just paste this snippet into #254 or something for tracking 👌

Issue #248, opened 2025-02-13
Minor language tweaks to x-parser error language


klebba:
It's awesome that these are all wrapped up in this single lookup — thanks for that! I'll turn this into a PR soon... in the meantime I'm fiddling here in the ticket for some unknown reason

```
A: ['#100', 'Could not parse markup at beginning of template.'],
B: ['#100', 'Markup at the start of your template could not be parsed.'],

A: ['#101', 'Could not parse markup after text content in template.'],
B: ['#101', 'Markup after content text found in your template could not be parsed.'],

A: ['#102', 'Could not parse markup after comment in template.'],
B: ['#102', 'Markup after a comment found in your template could not be parsed.'],

A: ['#103', 'Could not parse markup after interpolated content in template.'],
B: ['#103', 'Markup after a content interpolation in your template could not be parsed.'],

['#104', 'Markup after the tag name in an opening tag in your template could not be parsed.'],
['#104', 'Markup after the tag name in an opening tag in your template could not be parsed.'],

['#105', 'Markup after a spacing in an opening tag in your template could not be parsed.'],
['#105', 'Markup after a spacing in an opening tag in your template could not be parsed.'],

['#106', 'Markup after an opening tag in your template could not be parsed.'],
['#106', 'Markup after an opening tag in your template could not be parsed.'],

['#107', 'Markup after a boolean attribute in an opening tag in your template could not be parsed.'],
['#107', 'Markup after a boolean attribute in an opening tag in your template could not be parsed.'],

['#108', 'Markup after an attribute in an opening tag in your template could not be parsed.'],
['#108', 'Markup after an attribute in an opening tag in your template could not be parsed.'],

['#109', 'Markup after a boolean attribute interpolation in an opening tag in your template could not be parsed.'],
['#109', 'Markup after a boolean attribute interpolation in an opening tag in your template could not be parsed.'],

['#110', 'Markup after a defined attribute interpolation in an opening tag in your template could not be parsed.'],
['#110', 'Markup after a defined attribute interpolation in an opening tag in your template could not be parsed.'],

['#111', 'Markup after an attribute interpolation in an opening tag in your template could not be parsed.'],
['#111', 'Markup after an attribute interpolation in an opening tag in your template could not be parsed.'],

['#112', 'Markup after a property interpolation in an opening tag in your template could not be parsed.'],
['#112', 'Markup after a property interpolation in an opening tag in your template could not be parsed.'],

[...]

PR #283, opened 2025-03-06
Improve parser error messaging and factoring.

theengineear:
Taking on some of the language tweaks from #248. Also `switch`-ing on the values (versus using a `Map`) so that these cases can be enforced by code coverage.

We likely have a couple messages which are impossible to get to. It’d be good to be forced to either remove them or find a way to prove that they are reachable via our test suite.

For now, I reduced the coverage target to keep this change set smaller, but I will address that in follow-up efforts.


doc/TEMPLATES.md
```diff
@@ -291,6 +291,49 @@ html`<div>${fragment}</div>`;
 // Error: Unexpected child element count of zero for given DocumentFragment.
 ```
 
+## Supported native tags
```
theengineear, line 294: As requested, added a section for `#supported-native-tags` in our `TEMPLATES.md` file (so that we can link to it from our error messaging).

doc/TEMPLATES.md
```diff
@@ -291,6 +291,49 @@ html`<div>${fragment}</div>`;
...
+
+The default template engine is opinionated about which elements are allowed in
+order to reduce complexity and improve performance. The following tags are
+supported:
+
+```html
```
theengineear, line 300: Thought it would be nice to present this with `html` since that’s what it’s about. Open to different formatting options, but this felt like it would get the point across in a concise way.

test/index.js
```diff
@@ -7,7 +7,7 @@ import '../x-template.js';
 
 // Set a high bar for code coverage!
 coverage(new URL('../x-element.js', import.meta.url).href, 100);
-coverage(new URL('../x-parser.js', import.meta.url).href, 100);
+coverage(new URL('../x-parser.js', import.meta.url).href, 97); // TODO: Get back to 100%!
```
theengineear, line 10: Will fix this later (pinky promise).

test/test-parser.js
```diff
@@ -699,6 +699,42 @@ describe('odds and ends', () => {
 });
 
 describe('errors', () => {
+  it('throws when initial markup cannot be parsed', () => {
```
theengineear, line 702: Here are a couple cases I _was_ able to add to improve our coverage.

x-parser.js
```diff
@@ -228,99 +228,112 @@ export class XParser {
...
-    ['#151', 'Malformed hexadecimal character reference (html entity) or ambiguous ampersand.'],
-    ['#152', 'Malformed html comment. Comments cannot start with “>” or “->” characters, they cannot include a set of “--” characters, and they cannot end with a “-” character.'],
-    ['#153', 'Forbidden html element used — this parser is opinionated about which elements are allowed in order to reduce complexity and improve performance.'],
-    ['#154', 'Unmatched closing tag at the end of your template. To avoid unintended markup, non-void tags must explicitly be closed.'],
-    ['#155', 'Mismatched closing tag used. To avoid unintended markup, non-void tags must explicitly be closed and all closing tag names must be a case-sensitive match.'],
-    ['#156', 'Forbidden, nontrivial interpolation of <textarea> tag used. Only basic interpolation is allowed — e.g., <textarea>${…}</textarea>.'],
-    ['#157', 'Forbidden declarative shadow root used (e.g., `<template shadowrootmode="open">`).'],
-  ]);
+  static #getErrorMessage(key) {
+    switch (key) {
```
theengineear, line 237: If you `switch` on this stuff, this will all get enforced by our code coverage targets. I had initially started trying to duplicate this mapping and enforce it in our tests… but that’s brittle. It’s really nice that this will just _always_ be enforced moving forward.

x-parser.js
```diff
@@ -228,99 +228,112 @@ export class XParser {
...
+      case '#100': return 'Could not parse template markup (at template start).';
+      case '#101': return 'Could not parse template markup (after text content).';
```
theengineear, line 239: Took on most of the language tweaks. Thanks @klebba!

x-parser.js
```diff
@@ -519,15 +532,15 @@ export class XParser {
     const { parsed, notParsed } = XParser.#getErrorInfo(strings, stringsIndex, string, stringIndex);
     const valueForbidden = XParser.#forbiddenTransition(string, stringIndex, value);
     const valueMalformed = XParser.#invalidTransition(string, stringIndex, value);
-    const errorMessagesKey = valueForbidden
-      ? XParser.#valueForbiddenToErrorMessagesKey.get(valueForbidden)
+    const errorMessageKey = valueForbidden
```
theengineear, line 535: Refactored some of the variable names to match the new setup. Unfortunately, that made this diff a little extra ugly.

theengineear:
FYI @klebba — this change set doesn’t close down #248, but it should be a good step forward.

PR #284, opened 2025-03-06
Restore test coverage for “x-parser.js”.

theengineear:
In a prior change set, we changed our `Map` error mappings to `switch` error mappings so that our code coverage tests would ensure that we are able to prove that each error is possible to be shown — otherwise, why keep them.

This change set was kept separate since it’s all about coverage.

Note that some of the regular expressions have changed — nearly all of that is due to rewriting like `\s\n` as just `\s` since `\n` is one of the characters matched by the `\s` set.


test/index.js
```diff
@@ -7,7 +7,7 @@ import '../x-template.js';
 
 // Set a high bar for code coverage!
 coverage(new URL('../x-element.js', import.meta.url).href, 100);
-coverage(new URL('../x-parser.js', import.meta.url).href, 97); // TODO: Get back to 100%!
+coverage(new URL('../x-parser.js', import.meta.url).href, 100);
```
theengineear, line 10: And we’re back 🎉

x-parser.js
```diff
@@ -99,12 +99,10 @@ export class XParser {
   // Examples:
   //  - ok: <h6>, <my-element-1>
   //  - not ok: <-div>, <1-my-element>
-  static #startTagOpen            =   /<(?![0-9-])[a-z0-9-]+(?<!-)(?=[\s\n>])/y;
+  static #startTagOpen            =   /<(?![0-9-])[a-z0-9-]+(?<!-)(?=[\s>])/y;
   static #endTag                  = /<\/(?![0-9-])[a-z0-9-]+(?<!-)>/y;
-  static #startTagClose           =    /(?<![\s\n])>/y;
```
theengineear, line 104: There’s no reason to write `[\s\n]` — `\s` includes `\n`.

x-parser.js
```diff
@@ -113,7 +111,7 @@ export class XParser {
   // Examples:
   //  - ok: <div foo bar>, <div\n  foo\n  bar>
   //  - not ok: <div foo  bar>, <div\n\n  foo\n\n  bar>, <div\tfoo\tbar>
-  static #startTagSpace           = / (?! )|\n *(?!\n)(?=[-_.?a-zA-Z0-9>])/y;
```
theengineear, line 116: I was able to simplify this by adding a new _invalid state transition_ — namely, we now check to see if a space is followed by an invalid (or duplicate) space. Strictly-speaking, removing the lookahead _should_ eek out minor performance gains as well (although I didn’t scrutinize that this round).

x-parser.js
```diff
@@ -239,61 +237,57 @@ export class XParser {
       case '#101': return 'Could not parse template markup (after text content).';
       case '#102': return 'Could not parse template markup (after a comment).';
       case '#103': return 'Could not parse template markup (after interpolated content).';
-      case '#104': return 'Could not parse template markup (after a start tag name).';
```
theengineear, line 242: Because some of these were deemed _impossible_ to hit — I removed them and did some reordering of the error keys.

x-parser.js
```diff
@@ -382,6 +377,7 @@ export class XParser {
       case XParser.#danglingQuote: return XParser.#try(string, stringIndex,
         XParser.#startTagSpaceMalformed);
       case XParser.#startTagSpace: return XParser.#try(string, stringIndex,
+        XParser.#startTagSpaceMalformed,
```
theengineear, line 380: Again, this is the new _invalid_ state transition that was added to decrease the complexity of the `#startTagSpace` pattern.

x-parser.js
```diff
@@ -569,9 +565,20 @@ export class XParser {
...
+      case XParser.#boundProperty: {
+        const errorMessageKey = XParser.#getErrorMessageKeyFromErrorName('missing-closing-quote');
+        const errorMessage = XParser.#getErrorMessage(errorMessageKey);
+        const substringMessage = `Missing closing double-quote.`;
+        throw new Error(`[${errorMessageKey}] ${errorMessage}\n${substringMessage}`);
+      }
+    }
     if (tagName) {
-      const errorMessageKey = XParser.#getErrorMessageKeyFromErrorName('missing-closing-tag');
+      const errorMessageKey = XParser.#getErrorMessageKeyFromErrorName('missing-end-tag');
```
theengineear, line 581: Making our language consistent — missed this one last round!

test/test-parser.js
```diff
@@ -1419,6 +1419,181 @@ describe('errors', () => {
 
   it('throws for declarative shadow roots', () => {
     const callback = () => html`<template shadowrootmode="open"></template>`;
+    const expectedMessage = '[#156]';
+    assertThrows(callback, expectedMessage, { startsWith: true });
+  });
+});
+
+// Sanity check to make sure we’re able to hit all our errors.
+describe('errors coverage', () => {
```
theengineear, line 1428: Added a dedicated `errors coverage` test section where we can add odds-and-ends tests which are required to meet our 100% coverage goals.

x-parser.js
```diff
@@ -239,61 +237,57 @@ export class XParser {
...
 
   // Block #100-#119 — Invalid transition errors.
   static #getErrorMessageKeyFromValue(value) {
+    // Note that the following states _never_ have their own generic errors. The
+    //  main reason is because they all encode a lookahead which is a subset of
+    //  a related “*Malformed” pattern.
+    // - XParser.#startTagOpen  (caught by XParser.#startTagSpaceMalformed)
+    // - XParser.#boolean       (caught by XParser.#startTagSpaceMalformed)
+    // - XParser.#attribute     (caught by XParser.#startTagSpaceMalformed)
+    // - XParser.#danglingQuote (caught by XParser.#startTagSpaceMalformed)
```
theengineear, line 278: Added commentary as to why the list below doesn’t include _all_ the `value` states. Kinda neat that we were able to get our code coverage tool to find unreachable cases for us 🎉

PR #285, opened 2025-03-06
Adds arrays to performance test templates.

theengineear:
Only demo / test changes here — it’s not too hard to support this since most (all?) templating engines will accept a native _array_ as input for a set of children.

### DIFF
--- a/demo/performance/common.js
+++ b/demo/performance/common.js
@@ -58,6 +58,7 @@ export default class CommonTest {
       title: 'test',
       content1: 'AAA',
       content2: 'BBB',
+      items: [{ text: 'one' }, { text: 'two' }, { text: 'three' }],
     },
     {
       attr: '456',
@@ -76,6 +77,7 @@ export default class CommonTest {
       title: 'test',
       content1: 'ZZZ',
       content2: 'BBB',
+      items: [{ text: 'one' }, { text: 'two' }, { text: 'three' }, { text: 'four' }, { text: 'five' }, { text: 'six' }],
     },
   ];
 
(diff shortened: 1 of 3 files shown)


demo/performance/common.js
```diff
@@ -235,14 +237,19 @@ export class HtmlLiteralInterface {
...
+    const { attr, one, two, three, four, five, six, seven, eight, nine, ten, id, hidden, title, content1, content2, items } = properties;
     // eslint-disable-next-line no-eval
     return eval(`html\`<div data-id="p1" attr="\${attr}">
       <div data-id="p2" data-foo one="\${one}" two="\${two}" three="\${three}" four="\${four}" five="\${five}" .six="\${six}" .seven="\${seven}" .eight="\${eight}" .nine="\${nine}" .ten="\${ten}">
         <div data-id="p3" data-bar="bar">
           <div data-id="\${id}" boolean ?hidden="\${hidden}" .title="\${title}">
             \${content1} -- \${content2}
           </div>
+          <ul data-id="list">
+            \${(items ?? []).map(item => {
```
theengineear, line 249: Easy peasy to test / compare performance of this stuff — we might as well.

test/test-parser.js
```diff
@@ -1544,12 +1544,16 @@ describe('errors coverage', () => {
     assertThrows(callback, expectedMessage, { startsWith: true });
   });
 
+  //////////////////////////////////////////////////////////////////////////////
```
theengineear, line 1547: Missed adding these error-code-block delimiters to the parser tests before. Sneaking them in.

PR #287, opened 2025-03-09
Pass standard “Array.map” arguments to “repeat”.

theengineear:
The “repeat” directive is just a shim to create arrays of _entries_. We may remove this from the interface in the future, but we do need it to function as expected in the interim!

Most callers rely on just the first “value” argument, but there’s no good reason not to expose the `(value, index, array)`-triple that a standard callback would receive from “Array.map”.

### DIFF
--- a/test/test-template-engine.js
+++ b/test/test-template-engine.js
@@ -1241,6 +1241,38 @@ describe('html updaters', () => {
     assert(container.querySelector('#c').textContent === 'fuzz');
     container.remove();
   });
+
+  it('repeat callbacks are provided args from underlying “.map” call', () => {
+    // The identify callback is written in a bizarre way to test that all the
+    //  expected function arguments are actually passed in. If they weren’t you
+    //  would get duplicated key errors or undefined key errors.
+    const getTemplate = ({ items }) => {
+      return html`
+        <div id="target">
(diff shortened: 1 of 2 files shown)


test/test-template-engine.js
```diff
@@ -1241,6 +1241,38 @@ describe('html updaters', () => {
     assert(container.querySelector('#c').textContent === 'fuzz');
     container.remove();
   });
+
+  it('repeat callbacks are provided args from underlying “.map” call', () => {
```
theengineear, line 1245: FYI —Added this test _first_ to prove that we could repro failure in our test suite. Then, added the related fix.

x-template.js
```diff
@@ -126,8 +126,8 @@ class TemplateEngine {
       throw new Error(`Unexpected repeat callback "${callback}" provided, expected a function.`);
     }
     return identify
-      ? items.map(item => [identify(item), callback(item)])
-      : items.map(item => callback(item)); // Just a basic array.
+      ? items.map((...args) => [identify(...args), callback(...args)])
```
theengineear, line 129: Super simple fix to just pass everything through as-is.

theengineear:
FYI @klebba — noticed this one while I was working today. It was a quick fix, so I figured I’d throw in a quick update.

PR #289, opened 2025-03-19
Lock down CSP in “.html” files to prove we can.

theengineear:
We want integrators to be able to enforce the strictest content security policies possible — this change set places strict CSPs in our own documents to keep us honest.

### DIFF
--- a/demo/lit-html/demo-lit-html.js
+++ b/demo/lit-html/demo-lit-html.js
@@ -1,6 +1,17 @@
 import BaseElement from './base-element.js';
 
 export default class DemoLitHtml extends BaseElement {
+  static get styles() {
+    const styleSheet = new CSSStyleSheet();
+    styleSheet.replaceSync(`
+      #container[emoji]::before {
+        content: " " attr(emoji);
+        font-size: 2rem;
+      }
+    `);
+    return [styleSheet];
+  }
+
   static get properties() {
     return {
       emoji: {
@@ -16,12 +27,6 @@ export default class DemoLitHtml extends BaseElement {
   static template(html, { ifDefined }) {
     return ({ emoji, message }) => {
       return html`
-        <style>
-          #container[emoji]::before {
-            content: " " attr(emoji);
-            font-size: 2rem;
-          }
-        </style>
         <div id="container" emoji="${ifDefined(emoji)}">Rendered &ldquo;${message}&rdquo; using <code>lit-html</code>.</div>
       `;
     };
--- a/demo/lit-html/index.html
+++ b/demo/lit-html/index.html
@@ -1,7 +1,8 @@
 <!doctype html>
 <html>
   <head>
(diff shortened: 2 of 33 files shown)


demo/lit-html/demo-lit-html.js
```diff
@@ -16,12 +27,6 @@ export default class DemoLitHtml extends BaseElement {
   static template(html, { ifDefined }) {
     return ({ emoji, message }) => {
       return html`
-        <style>
```
theengineear, line 19: Simply changing this inline `<style>` tag usage so we can have a stricter content security policy on this particular _demo_. To be clear, this is not production code.

demo/uhtml/demo-uhtml.js
```diff
@@ -16,12 +27,6 @@ export default class DemoUhtml extends BaseElement {
   static template(html) {
     return ({ emoji, message }) => {
       return html`
-        <style>
```
theengineear, line 19: Simply changing this inline `<style>` tag usage so we can have a stricter content security policy on this particular _demo_. To be clear, this is not production code.

demo/performance/react.html
```diff
@@ -1,7 +1,12 @@
 <!doctype html>
 <html>
   <head>
-    <meta charset="UTF-8">
+    <meta charset="utf-8">
+    <!-- TODO: The react library has a bunch of dependencies which makes the CSP
```
theengineear, line 5: Where we were forces to loosen our policy _more_ than we’d like… I added a `TODO` comment. Here, for example, it’s brutal to try and enumerate all the individual `esm.sh` urls for transitive dependencies of React. So, I just allow any package there via `https://esm.sh/`. This is just a demo, so I’m not terribly worried about it, but I think it’s good to call out when integration is cumbersome.

By contrast, the `lit-html` policy was actually a breeze as it doesn’t _have_ dependencies (at least not for our usage in the demo).

theengineear:
The most important lines in this change set are the CSP `meta` tags applied to our `demo/index.html` and our `test/*.html` files — they prove that integrators can lock down CSP without `x-element` getting in the way.

theengineear:
FYI @klebba — Happily (and perhaps unsurprisingly), I was able to lock down policies in our documents without having to make any changes to the the production code.

demo/lit-html/index.html
```diff
@@ -1,7 +1,8 @@
 <!doctype html>
 <html>
   <head>
-    <meta charset="UTF-8">
+    <meta charset="utf-8">
+    <meta http-equiv="content-security-policy" content="default-src 'self'; script-src 'self' 'sha256-vpd4e6Q2EYlXOAcBOCPYNRqOSK1caV1iPamby7fBE30=' https://esm.sh/lit-html@3.2.1/;">
```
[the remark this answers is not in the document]
klebba: that's actually horribleawesome

PR #292, opened 2025-03-25
Update TEMPLATES.md

klebba:
Simplifying template map example to something more minimal, eliminating two superfluous callbacks


theengineear (approved):
Sure — works for me 👌

PR #304, opened 2025-06-27
Fix rendering lists of changing length.

### LINKED ISSUE #303: Re-rendering lists-of-lists where outer list resets to zero length breaks.
The following test proves this out — it will error:

```js
  it('renders lists of lists', () => {
    const getTemplate = ({ items }) => {
      return html`<div id="target">${items.map(item => {
        return html`${item.items.map(subItem => {
          return html`<div>${`:${item.text}-${subItem.text}:`}</div>`;
        })}`;
      })}</div>`;
    };
    const container = document.createElement('div');
    render(container, getTemplate({
      items: [
        { text: 'foo', items: [{ text: 'one' }] },
      ],
    }));
    assert(container.querySelector('#target').textContent === ':foo-one:');
    render(container, getTemplate({
      items: [], // Tests a specific issue we had going from some > none > some.
    }));
    assert(container.querySelector('#target').textContent === '');
    render(container, getTemplate({
      items: [
        { text: 'foo', items: [{ text: 'one' }] },
      ],
    }));
    assert(container.querySelector('#target').textContent === ':foo-one:');
  });
```

This issue can happen frequently in real-world apps where lists get defaulted back to `[]` when data is re-fetched from APIs.

theengineear:
The root cause of the error here was that we were inadvertently removing
our _outer_ cursor when removing DOM nodes through our inner cursors.

Rather than _remove through_, we need to _remove between_ and then also
remove our one additional inner cursor.

Additionally, we needed to do more to clean up some internal state when
lists are truncated.

Closes #303.

### DIFF
--- a/test/test-template-engine.js
+++ b/test/test-template-engine.js
@@ -459,10 +459,40 @@ describe('html rendering', () => {
       `;
     };
     const container = document.createElement('div');
-    render(container, getTemplate({ items: [1, 2, 3] }));
-    assert(container.querySelector('#target').childElementCount === 3);
-    render(container, getTemplate({ items: [1, 2, 3, 4, 5] }));
-    assert(container.querySelector('#target').childElementCount === 5);
+
+    // Render 0 > 1 > 2
+    render(container, getTemplate({ items: [] }));
+    assert(container.querySelector('#target').childElementCount === 0);
+    render(container, getTemplate({ items: [1] }));
+    assert(container.querySelector('#target').childElementCount === 1);
+    render(container, getTemplate({ items: [1, 2] }));
+    assert(container.querySelector('#target').childElementCount === 2);
+
+    // Render 2 > 1 > 0
+    render(container, getTemplate({ items: [1, 2] }));
+    assert(container.querySelector('#target').childElementCount === 2);
+    render(container, getTemplate({ items: [1] }));
+    assert(container.querySelector('#target').childElementCount === 1);
+    render(container, getTemplate({ items: [] }));
+    assert(container.querySelector('#target').childElementCount === 0);
+
+    // Render 0 > 1 > 0 > 1
+    render(container, getTemplate({ items: [] }));
+    assert(container.querySelector('#target').childElementCount === 0);
+    render(container, getTemplate({ items: [1] }));
+    assert(container.querySelector('#target').childElementCount === 1);
+    render(container, getTemplate({ items: [] }));
+    assert(container.querySelector('#target').childElementCount === 0);
+    render(container, getTemplate({ items: [1] }));
+    assert(container.querySelector('#target').childElementCount === 1);
+
+    // Render 0 > 2 > 0 > 2
+    render(container, getTemplate({ items: [] }));
+    assert(container.querySelector('#target').childElementCount === 0);
+    render(container, getTemplate({ items: [1, 2] }));
+    assert(container.querySelector('#target').childElementCount === 2);
+    render(container, getTemplate({ items: [] }));
+    assert(container.querySelector('#target').childElementCount === 0);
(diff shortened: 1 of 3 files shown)


test/test-template-engine.js
```diff
@@ -482,6 +482,16 @@ describe('html rendering', () => {
       ],
     }));
     assert(container.querySelector('#target').textContent === ':foo-one:');
+    render(container, getTemplate({
+      items: [], // Tests a specific issue we had going from some > none > some.
```
theengineear, line 486: Specifically flipping back to the _zero length_ case causes this issue to happen.

test/test-template-engine.js
```diff
@@ -459,10 +459,40 @@ describe('html rendering', () => {
...
+    assert(container.querySelector('#target').childElementCount === 1);
+    render(container, getTemplate({ items: [1, 2] }));
+    assert(container.querySelector('#target').childElementCount === 2);
+
+    // Render 2 > 1 > 0
+    render(container, getTemplate({ items: [1, 2] }));
+    assert(container.querySelector('#target').childElementCount === 2);
+    render(container, getTemplate({ items: [1] }));
+    assert(container.querySelector('#target').childElementCount === 1);
+    render(container, getTemplate({ items: [] }));
```
theengineear, line 476: Specifically flipping back to the zero length case causes one of the issues.

test/test-template-engine.js
```diff
@@ -459,10 +459,40 @@ describe('html rendering', () => {
...
+    assert(container.querySelector('#target').childElementCount === 0);
+
+    // Render 0 > 1 > 0 > 1
+    render(container, getTemplate({ items: [] }));
+    assert(container.querySelector('#target').childElementCount === 0);
+    render(container, getTemplate({ items: [1] }));
+    assert(container.querySelector('#target').childElementCount === 1);
+    render(container, getTemplate({ items: [] }));
```
theengineear, line 484: Specifically flipping back-and-fourth to the zero length case causes one of the issues.

test/test-template-engine.js
```diff
@@ -459,10 +459,40 @@ describe('html rendering', () => {
...
+    assert(container.querySelector('#target').childElementCount === 0);
+    render(container, getTemplate({ items: [1] }));
+    assert(container.querySelector('#target').childElementCount === 1);
+
+    // Render 0 > 2 > 0 > 2
+    render(container, getTemplate({ items: [] }));
+    assert(container.querySelector('#target').childElementCount === 0);
+    render(container, getTemplate({ items: [1, 2] }));
+    assert(container.querySelector('#target').childElementCount === 2);
+    render(container, getTemplate({ items: [] }));
```
theengineear, line 494: And — the other issue can be tested by going back-and-forth from _two_ (as it causes a specific keying issue).

x-template.js
```diff
@@ -470,8 +470,12 @@ class TemplateEngine {
       const index = values.length;
       const id = String(index);
       const item = arrayState.map.get(id);
-      TemplateEngine.#removeThrough(item.startNode, node);
-      arrayState.map.delete(id);
+      TemplateEngine.#removeBetween(item.startNode, node);
+      item.startNode.remove();
+      for (let iii = arrayState.map.size - 1; iii >= values.length; iii--) {
+        // We iterate backwards since we are deleting keys from the map itself.
+        arrayState.map.delete(String(iii));
```
theengineear, line 477: And, we need to make sure to property truncate the _map_. The alternative would be to keep an `array` and splice it to remove values… I’m not actually convinced that would be an overall performance win though since you are then forced to use array-access elsewhere.

Anyways, this was the simplest fix I could implement and we can do another performance pass in the future. I just really wanted to lock this stuff down in our test suite.

theengineear:
FYI @klebba 🤙

PR #305, opened 2025-06-27
Add “CHANGELOG.md” note for fix to array renders.

theengineear:


theengineear:
Minor follow on — forgot to commit the related change log entry.

PR #307, opened 2025-06-30
Remove “ready.js” ahead of “2.x” release.

### LINKED ISSUE #302: Drop ready.js in 2.0 stable
Migration to the 2.0 release will require developers depending on `etc/ready.js` to crib the file into their project source. The code is trivial:

```
/**
 * Await document completeness. Will likely be replaced by built-in apis.
 * Check out https://github.com/Netflix/x-element/issues/65 for details.
 * @param {Document} target
 * @returns {Promise<any>}
 */
const ready = target => {
  return new Promise(resolve => {
    if (target.readyState === 'complete') {
      resolve(target);
    } else {
      const handle = evt => {
        if (evt.target.readyState === 'complete') {
          target.removeEventListener('readystatechange', handle);
          resolve(target);
        }
      };
      target.addEventListener('readystatechange', handle);
    }
  });
};

export default ready;
```

As browsers adopt CSS modules via `with { type: css }` import attributes, the need to use a DOMReady guard to skirt the Flash of Unstyled Content (FoUC) will finally pass. Instead we will use `blocking="render"` on imported custom elements. This allows the browser to wait until both JS and CSS content has loaded.

theengineear:
Native APIs will fully replace the functionality offered by this file. As such, we will prune it ahead of the “2.x” release.

In particular, the `blocking="render"` attribute / value pair spec has been merged in — see https://github.com/whatwg/html/pull/7474.

Closes #302.


theengineear:
@klebba — `ready` for this to go?

PR #312, opened 2025-08-05
Allow binding empty “DocumentFragment” instances.

### LINKED ISSUE #310: Document fragment child element count guard not working as intended
See:
https://github.com/Netflix/x-element/blob/b734ddcffb37eb0ffd1f828f25013d1187232759/x-template.js#L585-L587

Image

A document fragment may have non-element child nodes; we should relax this guard

theengineear:
Previously, we had a small developer ergonomics guard to throw if an empty document fragment was bound as content. This commit reverts that prior decision for two reasons:

1. The guard checked for “element” children (not “node” children) as we internally use comment node cursors to annotate our DOM. This was too naive as it also doesn’t detect _text nodes_.
2. The guard was misaligned with our philosophy — we don’t want to teach developers how the browser works. I.e., we expect our users to be experts and so we don’t want to inject special decisions which would infer intent and potentially frustrate expert users.

Closes #310.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - There are now three files in this repo — “x-element.js” which is all about
   element interfaces, “x-template.js” which is all about DOM management, and
   “x-parser.js” which is all about interpolated html markup interpretation.
+- An error is no longer thrown when binding an empty document fragment as
+  content in a template (#310).
 
 ### Removed
 
--- a/doc/TEMPLATES.md
+++ b/doc/TEMPLATES.md
@@ -275,20 +275,15 @@ html`<div>${bar}</div>`;
 When the content being bound is a `DocumentFragment` (e.g., from a `<template>`
 element’s `.content` property), the child nodes of that fragment will be added
 via an `.append(fragment)` on the parent container. This _moves_ those nodes
-from the fragment to the container (i.e., _not_ a copy). To prevent developers
-from unintentionally re-binding an already-exhausted fragment, the template
-engine _will throw_ when an empty document fragment is provided.
+from the fragment to the container (i.e., _not_ a copy).
 
 ```js
 const template = document.createElement('template');
-template.setHTMLUnsafe(`<script>console.log('hello world');</script>`);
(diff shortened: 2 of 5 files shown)


CHANGELOG.md
```diff
@@ -15,6 +15,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - There are now three files in this repo — “x-element.js” which is all about
   element interfaces, “x-template.js” which is all about DOM management, and
   “x-parser.js” which is all about interpolated html markup interpretation.
+- An error is not longer thrown when binding an empty document fragment as
```
theengineear, line 18: This exists outside of the `rc` releases, so it’s worth nothing in the change log.

doc/TEMPLATES.md
```diff
@@ -275,9 +275,7 @@ html`<div>${bar}</div>`;
 When the content being bound is a `DocumentFragment` (e.g., from a `<template>`
 element’s `.content` property), the child nodes of that fragment will be added
 via an `.append(fragment)` on the parent container. This _moves_ those nodes
-from the fragment to the container (i.e., _not_ a copy). To prevent developers
-from unintentionally re-binding an already-exhausted fragment, the template
-engine _will throw_ when an empty document fragment is provided.
```
theengineear, line 280: Updated documentation to reflect that this no longer throws.

theengineear:
@klebba — I’ll hold off on merging so you can take a look. I believe this is what we agreed on.

CHANGELOG.md
klebba: typo: `no longer thrown`

doc/TEMPLATES.md
```diff
@@ -286,9 +284,6 @@ const fragment = template.content.cloneNode(true);
 
 html`<div>${fragment}</div>`;
 // <div><script>console.log('hello world');</script></div>
```
klebba, line 286: Can we update this example to avoid this antipattern of inline script? IMO this is not the use case we want to exemplify

klebba (approved):
LG2M — left a few comments to consider. Feel free to merge on my prior

doc/TEMPLATES.md
klebba: After our discussion: svg makes more sense because it has children nodies

Issue #314, opened 2025-08-06
2.0 pre and post-release work


klebba:
Tasks to complete pre-release

*  ✅ Remove support for `ifDefined` and `repeat` directives (and associated tests, docs, etc) — [updated 2025-12-03] implemented in #335.
* ✅ Remove support for custom template engines (and associated tests, demos, etc) — [updated 2025-12-03] implemented in #338.
*  ✅ Ensure properties exist on the instance at construction time (React interop enhancement, also forward looking toward future class decorator support) — [updated 2025-12-03] implemented in #332.
*  ✅ Decide: remove support for `ifDefined` and `repeat`, or continue to support as deprecated features (and all related work, e.g. removing comments and other complexity) — [updated 2025-12-02] team decides to remove support completely in 2.0
*  ✅ Decide: should we drop support for custom template engines? — [updated 2025-12-02] team decides to remove support completely in 2.0
*  ✅ Decide: what happens when you call "render" before `onConnected` (throw vs do it) — [updated 2025-12-02] team decides to fix the bug (do it, do not throw)

Tasks to complete post-release — #215 

* Fix the render-before-onConnected lifecycle woe
* Symbol.toStringTag implementation #317 
* Rethink definition-time analysis routine (versus lazy time): we are doing a lot of heavy lifting during the "import waterfall" / and testing elements uncovers some "code smell" regarding when analysis happens - it would be preferable to isolate analysis from registration. We could do "lazy analysis" instead
* Decide: rename project to "custom elephant 🐘"?
*  ✅ Think about the future of `DocumentFragment` support (e.g. reuse of doc frags is a confusing thing if you have forgotten their intrinsic behavior - so uhhh... do not forget it) — [updated 2025-12-03] this work has actually already been discussed and implemented, see #312

Post-release housekeeping
* review docs; simplify, clarify
* review error messages; we have a list in #248
* provide more pragmatic examples in /docs/EXAMPLES.md or similar
* unleash Claude on the source
* review array and map validation (e.g., we could relax errors in some cases)

theengineear:
Yep, just working with folks to ensure that we are using the latest security standards for our release pipeline. Probably will be able to get it shipped in the next week — I’ll just keep doing updates on the margins 🤙

klebba:
I upgraded my various projects and so far have not caught any issues — noice

PR #316, opened 2025-08-07
Use idiomatic “queueMicrotask” feature.

theengineear:
This feature was added as part of ES2019 and carries better semantics. I.e., any developer can tell that the point of calling it is to literally queue a microtask — it’s called “queueMicrotask”.

Functionally, it should be the same as `await Promise.resolve()`, but less cryptic.


theengineear:
Small track of work related to #314.

x-element.js
```diff
@@ -918,9 +918,8 @@ export default class XElement extends HTMLElement {
         computeMap.get(property).valid = false;
       }
       if (queueUpdate) {
-        // Queue a microtask. Allows multiple, synchronous changes.
-        await Promise.resolve();
-        XElement.#updateHost(host);
+        // Batch on microtask to allow multiple, synchronous changes.
```
theengineear, line 921: I nearly just deleted the comment because it’s _nearly_ self-documenting. But I think it’s worth highlighting that this is all about _batching_.

ts/x-element.d.ts
```diff
@@ -179,7 +179,7 @@ export default class XElement extends HTMLElement {
     static "__#3@#getListener"(host: any, listener: any): any;
     static "__#3@#updateHost"(host: any): void;
     static "__#3@#toPathString"(host: any): string;
-    static "__#3@#invalidateProperty"(host: any, property: any): Promise<void>;
+    static "__#3@#invalidateProperty"(host: any, property: any): void;
```
theengineear, line 182: Bonus — it does truly make the internal interface simpler.

PR #323, opened 2025-08-14
Optimize DOM cleanup with replaceChildren() and Range.deleteContents()

teimurjan:
## Description
Refactored DOM-clearing logic to use modern, batch-optimized APIs for improved performance and readability.

## Changes
1. **`removeChild` loop → `replaceChildren()`**
   - Replaces manual reverse iteration with the spec’s optimized “replace all” algorithm.
   - **Benefits:**
     - **Fewer mutations:** Single `childList` record instead of N per child.
     - **Better performance:** Allows browsers to batch DOM operations internally.
     - **Clearer intent:** Reads as “replace all children” without manual index handling.

2. **`while(...).remove()` → `Range.deleteContents()`**
   - Replaces sibling-by-sibling removal with a single range deletion.
   - **Benefits:**
     - **Bulk removal in one operation:** Reduces DOM tree surgery and mutation observer churn.
     - **Optimized in all major engines:** Common internal path used by browser editing commands.
     - **Simpler code:** Explicitly expresses “remove everything from start to end”.

## Before
```js
// Clear all children
const childNodes = node.childNodes;
for (let i = childNodes.length - 1; i >= 0; i--) {
  node.removeChild(childNodes[i]);
}

// Remove all siblings until startNode
while (node.previousSibling !== startNode) {
  node.previousSibling.remove();
}
```

## After
```js
// Clear all children
node.replaceChildren();

// Remove all siblings until startNode
const range = document.createRange();
range.setStartBefore(startNode);
range.setEndAfter(node);
range.deleteContents();
```

## Impact
- Improves performance for large DOM updates by reducing per-node mutations.
- Minimizes layout thrashing and MutationObserver overhead.
- Makes code more maintainable and self-explanatory.

### DIFF
--- a/x-template.js
+++ b/x-template.js
@@ -808,11 +808,7 @@ class TemplateEngine {
   //   }
   // }
   static #removeWithin(node) {
-    // Iterate backwards over the live node collection since we’re mutating it.
-    const childNodes = node.childNodes;
-    for (let iii = childNodes.length - 1; iii >= 0; iii--) {
-      node.removeChild(childNodes[iii]);
-    }
+    node.replaceChildren()
   }
 
   // TODO: Future state — we may choose to iterate differently as an


theengineear:
@teimurjan — Did I hear someone say “performance”?!

Thanks for checking out this project and triple-thanks for the PR ❤️ 

Here are a couple high-level notes from me (have only dug in a little bit):
* Performance tuning (as you likely know) is hard! Even harder to test… I love the more idiomatic approach you’ve taken, but I would like to see if that can be proved out.
* I have been thinking about ways to incorporate “ranges” and I wonder if a slightly different approach might be a better global optimization — one where we store the range as array metadata based on comment nodes (more on that below).

I took a moment to create some tests on jsbench. Here are my (very quick-and-dirty) findings:
* I think I’m convinced of the change to `removeChild loop → replaceChildren()` — here’s my [remove-within](https://jsbench.me/vqmfr5kzqk/1) test. Everything other than t he zero-children-case appears to be more performant. And, it’s obviously _much_ more idiomatic. I have a hazy memory of trying this long ago and getting _much_ different results — so this is an awesome win! It might even be worth just updating _this one method_ as its own change set.
* The benefits of the change to `while(...).remove() → Range.deleteContents()` seems maybe murkier? Here’s my [remove-between](https://jsbench.me/v9mfr8dgbr/1) test. That said, I _really_ like the idea of caching and maintaining references to ranges as part of our internal array metadata. Let me know if that resonates or is something you’d be interested in looking at!

I make no claims at all that my tests (or testing environment) are perfect, but I’m curious what you would think about trying to confirm / deny the findings there?

theengineear:
To be really clear about the “array state” comment. We hang some internal state off the `#ARRAY_STATE` symbol. I wonder if we could return a `range` or something from [createCursors](https://github.com/Netflix/x-element/blob/main/x-template.js#L771) and then we would just have to call `.deleteContents()` since the start / end cursors don’t change between re-renders of items.

teimurjan:
Hi @theengineear 👋

I am happy to help.

I've had a look at your benchmarks, and they seem solid.

1. I can confirm that testing `removeChild loop → replaceChildren()` indeed performs better in all the cases except zero children.
2. You're definitely correct regarding the Range API. I did deeper research, and `removeChild` happened to be much more optimized under the hood (especially for modern browsers). Range API does more work internally (validate ranges, handle partial text nodes, update live ranges, etc.). The creation of a Range is cheap. The significant impact is caused by setting the boundaries and deleting the content within these boundaries. That's why I doubt if having Range in `ARRAY_STATE` would help. It will introduce a micro-save on the creation only. The existing approach is an optimal one.

For now, let me revert the Range-related changes and keep `replaceChildren` only.

PR #328, opened 2025-11-01
Use “x-test-cli” test package.

theengineear:
The “x-test” ecosystem added a sibling tool called “x-test-cli” which exposes a CLI which can be added straight into NPM test scripts.

As part of the change, I noticed some NPM vulnerability warnings related to dev dependencies — so those were also updated.


doc/TESTING.md
```diff
@@ -11,3 +13,18 @@ npm test
...
+## Running specific tests
+
+You can filter tests by name using the `--test-name` argument:
+
+```bash
+# Run all tests matching a regex pattern
+npm test -- --test-name="basic"
+
+# Run tests from a specific test suite
+npm test -- --test-name="computed properties"
```
theengineear: This type of interface in node (and the query param interface in a browser) has proved incredibly helpful when working with AI. In particular, it allows agents to target specific issues one-by-one (versus having to think through the entire test suite en masse) — or try to `| grep` the output (which often doesn’t work well since it misses the errors).

The in-browser filtering ends up being really powerful for interopoeration with tools like the Chrome Devtools MCP server.

theengineear:
FYI @klebba — Trying out a couple changes to improve how AI can better-leverage large test suites. This has been working well for me on a couple projects and I thought it was worth pulling into this project as well. I’m going to get this merged (doesn’t impact interface / functionality at all), but we can chat at some point about the proposed `rc` interface for the `x-test` CLI tool.

PR #330, opened 2025-11-01
Bump “eslint-plugin-jsdoc” for stricter linting.

### LINKED ISSUE #329: Update “eslint-plugin-jsdoc” (stricter / different linting).
Just a chore issue to do some dependency management.

Image

theengineear:
This updates through some major versions of this dev dependency which causes many stricter rules in the recommended set.

Mostly, they are all about declaring types as `unknown` versus `any`, which is closer to what we _meant to say in the first place_.

The `any` type says — this is literally the union of everything that exists. The `unknown` type says — we know literally nothing about this thing.

Where the `any` type can be assigned to anything and anything can be assigned to it. The `unknown` type cannot be assigned to anything, but (importantly) anything can be assigned to it.

Closes #329.


x-template.js
```diff
@@ -94,8 +94,8 @@ class TemplateEngine {
    * html`<a href="${ifDefined(obj.href)}"></div>`;
    * ```
    * @deprecated
-   * @param {any} value
-   * @returns {any}
+   * @param {unknown} value
```
theengineear: Here, the semantics are now _much better_. It’s not that this is _any_, rather, we just say “we dunno what this is, so we’ll force ourselves to check.”

PR #332, opened 2025-11-19
Initialize properties in constructor for React.

### LINKED ISSUE #331: Consider “eagerly” defining props on prototype at construction time for React 19 interop.
React 19+ will perform a bit of _magic_ to try and figure out if it ought to set an _attribute_ or a _property_ when it encounters a binding on a regular html element (native or custom).

To do this magic, it does some introspection. However, we have been _very_ careful to not define properties until the custom element is actually _connected to the DOM_ (which is a sensible choice). This creates an issue though as React cannot properly introspect on the first render! This causes React to bind something like `myProp={{}}` as `myprop="[object Object]"` — yuck.

**Proposal**

Set all public properties to a special, internal Symbol _on construction_. This should be fairly cheap to do and would _roughly_ make elements appear to React as this:

```js
class MyElement extends XElement {
  someProp = XElement.#specialSymbol;

  static get properties() {
    return {
      someProp: {
        type: String,
      },
      /* … */
    };
  }

  /* … */
}
```

**Caveat**

Note that React will fail to “do the right thing” if the custom element happens to not be defined at the time of binding… that’s not in our control.

theengineear:
React doesn’t provide a way for developers to signal that they want to set a property (versus an attribute) in JSX — instead it performs some magic to _guess_ what was meant. This _guess_ is based on whether the expression `key in element` is `true`.

As a best-practice, XElement is _lazy_ about initialization and waits until it is _actually connected to the DOM_ before initializing all properties. But… that means React will always check for setters _before_ they exist!

This change set adds our accessor descriptors during construction, but adds an internal flag to prevent certain behavior until the component is fully initialized during connection.

Closes #331.

TODO:
- [x] Update React demo to prove this works. Ideally, we even write JSX and transform it so it all appears as idiomatic React.
- [x] Check in on performance tests. Hopefully this doesn’t cause an unexpected bottleneck!
- [x] Discuss how `delete` ought to work on elements… we have no prior tests on support for like `delete el.foo` in the pre-upgraded, pre-connected, and post-connected states. Might be worth considering what the “right” thing is here.
- [x] Test `delete` operator behavior and lock down expectations.
- [x] Update CHANGELOG.md.


x-element.js
```diff
@@ -667,6 +667,27 @@ export default class XElement extends HTMLElement {
...
+        if (Reflect.has(host, key)) {
+          valueMap.set(property, host[key]);
+        }
+      }
+    }
+    XElement.#upgradeOwnProperties(host);
+    for (const property of propertyMap.values()) {
+      if (!property.internal) {
+        const { key } = property;
+        Reflect.defineProperty(host, key, {
```
theengineear: Here’s the crux change — we make sure that there is a descriptor associated with each public property.

theengineear:
One other thought — it would be interesting to open up a PR for https://custom-elements-everywhere.com/ to add a test which checks that you can indeed set _properties_ on un-upgraded (or pre-connection) custom elements and have those correctly set properties on post-upgrade (or post-connection).

I.e., rather than trying to open up an issue against React, it’d be better to get a formal test added for an important edge-case in a custom-element lifecycle. I wonder if any other frameworks would fail that test 🤔

test/test-property-deletion.js
```diff
@@ -0,0 +1,63 @@
...
+  const el = document.createElement('test-delete-element');
+  const descriptor = Object.getOwnPropertyDescriptor(el, 'foo');
+  assert(descriptor !== undefined);
+  assert(descriptor.configurable === false);
+  assert(descriptor.enumerable === true);
+  assert(typeof descriptor.get === 'function');
+  assert(typeof descriptor.set === 'function');
+});
+
+it('prevents deletion of properties after construction', () => {
```
theengineear: Added some tests to lock down `delete` operator behavior. In this case — we now _throw_ if you try and call `delete` on a property before initialization. After a _lot_ of waffling, I think this is the correct behavior. The alternative is to silently let you delete which could cause silent bugs.

x-element.js
```diff
@@ -662,10 +662,48 @@ export default class XElement extends HTMLElement {
...
     }
-    XElement.#hosts.set(host, {
+    const hostInfo = {
       initialized: false, reflecting: false, invalidProperties, listenerMap,
       renderRoot, render, template, properties, internal, computeMap,
       observeMap, defaultMap, valueMap,
+    };
+    XElement.#hosts.set(host, hostInfo);
+    XElement.#upgradeOwnProperties(host);
+    // Define properties during construction for React 19 introspection (#331).
```
theengineear: This is the crux of the change. We need to define properties during _construction_ so React (and other frameworks) can reliably detect that they should set values as properties (not attributes).

x-element.js
```diff
@@ -662,10 +662,49 @@ export default class XElement extends HTMLElement {
...
+  static #defineProperty(host, property) {
+    const hostInfo = XElement.#hosts.get(host);
+    const { valueMap } = hostInfo;
+    const { key } = property;
+    Reflect.defineProperty(host, key, {
+      configurable: false,
+      enumerable: true,
+      get() {
+        const { initialized } = hostInfo;
+        if (initialized) {
```
theengineear: Again, we have an internal fork here… but I think this is perhaps the best we can do. The internal logic is _slightly_ different for setting / getting properties before initialization. Importantly, we just dump values to a _dumb_ value store pre-initialization and then do all the validation / initialization only after connection.

I went down a rabbit hole investigating whether we ought to try and _fully_ support properties post-construction / pre-initialization. However, things get weird pretty fast… you get into this strange middle-ground where something work _perfectly_ and then others… only somewhat. Most of that stems from rules around conformance for custom elements during construction. In particular, some of these rules exist because an element can be constructed before its attributes are fully parsed… this means you would have to deal with this middle-ground where you are trying to compute properties, sync programmatic attributes, reflect programmatic attributes, and render… but you _might_ be missing attributes which are yet-to-be-parsed in the markup.

Anyways… this is what I netted out with.

x-element.js
```diff
@@ -765,19 +804,22 @@ export default class XElement extends HTMLElement {
...
+    const { computeMap, initialized, invalidProperties, valueMap } = hostInfo;
     if (initialized === false) {
-      XElement.#upgradeOwnProperties(host);
       // Only reflect attributes when the element is connected.
       const { propertyMap } = XElement.#constructors.get(host.constructor);
       for (const property of propertyMap.values()) {
         const { value, found } = XElement.#getPreUpgradePropertyValue(host, property);
-        XElement.#initializeProperty(host, property);
         if (found) {
-          host[property.key] = property.default(host, property.initial(value));
```
theengineear: There were a couple options here:

1. Introduce a new flag like `initializing` (versus `initialized`) which could help us internally “do the right thing” when we _set_ a value on the host…
2. Use specific, internal methods during initialization because we accept that we are in a strange half-way state and it doesn’t really make sense to be using the full functionality _just yet_.

I went with (2) because the code felt cleaner overall. Additionally, by doing this, I was actually able to _remove _ an internal fork within `#invalidateProperty` 🤙

x-element.js
```diff
@@ -907,20 +931,18 @@ export default class XElement extends HTMLElement {
   }
 
   static #invalidateProperty(host, property) {
-    const { initialized, invalidProperties, computeMap } = XElement.#hosts.get(host);
-    if (initialized) {
```
theengineear: This was a big win — because we moved our forks higher up the call stack… we no longer need to check for `initialized` so deep in the call stack here. I never liked that we had to do this.

x-element.js
```diff
@@ -929,6 +951,18 @@ export default class XElement extends HTMLElement {
     return property.compute?.(host) ?? valueMap.get(property);
   }
 
+  static #validatePropertyMutable(host, property) {
```
theengineear: Had to factor this out so that we can call it surgically during initialization.

PR #333, opened 2025-11-19
Bump demo library versions.

theengineear:
We need to periodically update these in order to do an apples-to-apples comparison of the latest offerings for other template engines.

This also makes the following changes:
- Unifies import patterns in demos `import @netflix/x-element/*`.
- Makes React code idiomatic with `.jsx` support in demos.
- Unifies demo / performance test structure.

Note, React 19+ _hangs_ on the performance tests now, so that is skipped by default.


performance/default/index.js
```diff
@@ -1,5 +1,5 @@
-import { render, html } from '../../x-template.js';
-import CommonTest, { HtmlLiteralInterface } from './common.js';
+import { render, html } from '@netflix/x-element/x-template.js';
```
theengineear: I also added an import map to make these demo / performance imports “look more familiar”. I.e., from the demo’s perspective, we are now treating the dependency as “external” (even though it is indeed mapping to a local file within the import map).

performance/default/index.html
```diff
@@ -0,0 +1,22 @@
...
+    <!-- We use “unsafe-eval” below because we need to trick the browser into
+         not caching tagged template string objects to test interpretation. -->
+    <meta http-equiv="content-security-policy" content="
+      default-src 'self';
+      script-src 'self' 'unsafe-eval' 'sha256-3xomsitg9ZOvQzjbbIGFzD2Xto92sDJKO5NIlRHta9Q=';">
+    <link rel="stylesheet" href="../common.css">
+    <script type="importmap">
+      {
+        "imports": {
+          "@netflix/x-element/": "../../"
```
theengineear: Again — here’s where we allow the “internal” dependency to be treated as an “external” dependency in our demo code. I think this is a nice addition that properly demonstrates what it ought to look like to integrate the functionality.

theengineear:
This doesn’t impact functionality in any way, so I’m going to get these changes in. We can always revert if we end up changing our minds here.

PR #335, opened 2025-12-03
Ditch Deprecated Directives (e.g., “ifDefined”).

### LINKED ISSUE #336: Delete deprecated “ifDefined” and “repeat” directives.
Just creating a clear issue ticket for tracking. The original deprecation of these directives was discussed in #204. Then, we officially put them on the chopping block in #314. Rather than tie this specific need to the broader “2.x” release — I wanted to create a standalone ticket.

**The `ifDefined` Directive**

The `ifDefined` directive is replaced by a first-class DSL syntax — `??`. The following will only bind the `data-name` attribute _iff_ the value of `name` is not-nullish.

```html
<div ??data-name=${name}></div>
```

**The `repeat` Directive**

The `repeat` directive is replaces by idiomatic JS arrays. Specifically, the concept of “entries”. Conceptually, rather than provide a mapping function, the template engine allows you to pass “entries” like [Map.prototype.entries](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map/entries). As in:

```html
<div>
  ${deck.map(card => [card.id, html`<my-card id="${card.id}">${card.symbol}</my-card>`])}
</div>
```

theengineear:
* Removes “ifDefined” (previously just deprecated).
* Removes “repeat” (previously just deprecated).
* Removes some “future state” commentary (cleanup).

Closes #336.


theengineear:
@klebba — Dusting off this old commit and will get it merged momentarily. I added some extra issues to track and will update #314 accordingly.

PR #338, opened 2025-12-03
Remove custom template engine support.

### LINKED ISSUE #337: Remove support for custom template engines.
The `x-element` class currently provides the ability to plug in custom template engines as an alternative to the default `x-template` implementation. While this flexibility was well-intentioned, maintaining this pluggability creates an unsustainable API surface area. We cannot realistically guarantee that all template engines (current or future) will work correctly with `x-element`'s property system, lifecycle hooks, and DOM update mechanisms. Edge cases around property bindings, lifecycle timing, and diffing strategies vary significantly across implementations. Attempting to support all approaches means we either over-constrain our own evolution or create fragile integration points.

Moreover, the practical developer-facing value of this pluggability appears minimal. Developers who prefer other template engines can use them directly — custom elements work perfectly fine with any rendering library without requiring `x-element`'s base class. By removing this abstraction layer and tightening the integration between `x-element` and `x-template`, we can reduce maintenance burden, simplify the library, and enable future optimizations.

theengineear:
The ability to override `templateEngine` and use alternative rendering engines has been removed. The `x-element` class now uses `x-template` exclusively, which simplifies the codebase and reduces surface area.

Changes:
- Removes `defaultTemplateEngine` and `templateEngine` static getters.
- Updates `template()` method signature to remove `engine` parameter.
- Imports `html` and `render` directly from `x-template`.
- Removes related demos.
- Updates documentation to remove customization examples.
- Removes related tests.

Closes #337.


klebba:
NOICE

Issue #326, opened 2025-09-23
Review / Update GitHub Actions


theengineear:
Just jotting down some notes for later:

- [ ] Check on sub workflows like `actions/checkout` and `actions/setup-node` — we may be able to bump versions there.
- [ ] Migrate to environment-based tokens.
- [ ] Look into using additional authentication factors during publish.
- [ ] Look into using the [--provenance](https://docs.npmjs.com/generating-provenance-statements) flag when publishing, I believe that’s the best practice these days.

panagiotisbellias:
No problem at all — totally understand.

I first heard about x-element through GitHub while browsing open-source work related to UI component systems and was curious to explore the project further. I appreciate the approach and wanted to see if I could contribute in any small way.

Happy to follow along and help where it makes sense in the future!

PR #340, opened 2025-12-04
Use “npm-publish” environment to publish to NPM.

theengineear:
Doing this improves our security as it relates to publishing to NPM from GitHub. These changes relate to #326.

Additionally:
- Stop calling `npm ci` during publish, it’s a known attack vector and we shouldn’t actually need it in our no-build repository.
- Bump GitHub actions dependencies to “@v6”.

### DIFF
--- a/.github/workflows/publish-to-npm.yaml
+++ b/.github/workflows/publish-to-npm.yaml
@@ -5,13 +5,13 @@ on: [workflow_call]
 jobs:
   main:
     runs-on: ubuntu-latest
+    environment: npm-publish
     steps:
-      - uses: actions/checkout@v4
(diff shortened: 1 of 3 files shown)


.github/workflows/publish-to-npm.yaml
```diff
@@ -5,13 +5,13 @@ on: [workflow_call]
 jobs:
   main:
     runs-on: ubuntu-latest
+    environment: npm-publish
```
theengineear: **Key change**: We now run this action within an assumed environment (yet to be created). It’s called `npm-publish` here to differentiate between other ways we publish (e.g., JSR). I.e., `publish` did not feel sufficient enough.

PR #343, opened 2025-12-06
Use “moveBefore” for map shuffling.

### LINKED ISSUE #254: Neither connectedCallback nor disconnectedCallback should be called when re-mapping causes a shuffle.

theengineear:
Previously, both “connectedCallback” and “disconnectedCallback” would be called when simply attempting to reshuffle elements in a mapping within the same DOM tree.

With the addition of the “moveBefore” element API, we can fix this! Now, such children are simply _moved_ and remain connected.

Additionally, a new test was added to confirm that browsers are calling the “connectedMoveCallback” method on moves. And, another test added to ensure that we are testing all conditions / ternaries in our code to maintain 100% coverage.

Closes #254.

- [x] Works in Chrome
- [x] Works in FireFox (nightly)
- [ ] Works in Safari (see https://bugs.webkit.org/show_bug.cgi?id=281223)


test/test-scratch.js
```diff
@@ -103,6 +103,11 @@ class TestElement extends XElement {
     super.adoptedCallback();
     this.adopted = true;
   }
+
+  connectedMoveCallback() {
+    super.connectedMoveCallback();
```
theengineear: This is the only line in our test suite that exercises this callback — just for coverage.

test/test-template-engine.js
```diff
@@ -918,29 +976,28 @@ describe('html rendering', () => {
...
     assert(connects === 2);
-    // assert(moves === 0);
+    assert(moves === 0);
     assert(disconnects === 2);
 
     container.remove();
   });
 
-  // TODO: #254: See https://chromestatus.com/feature/5135990159835136.
-  it.todo('native map does not cause disconnectedCallback on list shuffle', () => {
```
theengineear: Very satisfying to delete this “.todo” here and see it _just work_ ❤️

x-element.js
```diff
@@ -153,11 +153,10 @@ export default class XElement extends HTMLElement {
...
 
-  // TODO: #254: Uncomment once we leverage “moveBefore”.
-  // /**
-  //  * Extends HTMLElement.prototype.connectedMoveCallback.
-  //  */
-  // connectedMoveCallback() {}
+  /**
+   * Extends HTMLElement.prototype.connectedMoveCallback.
+   */
+  connectedMoveCallback() {}
```
theengineear: Minor interface addition here for discoverability (matches our other patterns — i.e., you should be able to rely on `super` existing).

PR #355, opened 2026-03-06
Add a “test” skill to improve agentic workflows.

theengineear:
I have used this skill on other projects leveraging “x-test” and it has improved the efficiency of AI collaborations for me. At some point, we could consider documenting a skill like this in “x-test”, but there is some “x-element”-specific stuff in here now, and it feels premature to over-engineer something as simple as a skill.

With this, agents should be able to better interpret conversational requests to test and better-execute iterative workflows which involve both testing via the CLI and via a browser through tools like Chrome DevTools MCP.

### DIFF
--- a/.claude/skills/test/SKILL.md
+++ b/.claude/skills/test/SKILL.md
@@ -1,0 +1,48 @@
+---
+name: test
+description: Use when the user asks to run tests, check tests, verify tests
+             pass, or after making code changes that should be validated. Runs
+             unit tests either via CLI (puppeteer) or by visiting a test page
+             in the browser via Chrome DevTools MCP.
+---
+
+# Unit Tests
+
+Tests run in a real browser by loading a document (`test/test-parser.html`).
+Pick the narrowest URL that covers the change.
+
+- All tests: `http://localhost:8080/test/`
+- One test: `http://localhost:8080/test/test-parser.html`
+
+Output is [TAP v14](https://testanything.org/tap-version-14-specification.html).
+Preserve it or summarize plainly ("5/5 passed"). No emojis or checkmarks.
+
+## CLI
+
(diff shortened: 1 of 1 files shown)


.claude/skills/test/SKILL.md
```diff
@@ -0,0 +1,181 @@
...
+  - MCP: `http://localhost:8080/test/?x-test-name=click+events`
+
+  The pattern is a regex match, so partial matches work (e.g., "click events"
+  will match "handles click events").
+
+- **If working on a specific component**: Run only that component's tests
+
+- **If no specific context**: Run all tests
+
+Use judgment to pick the most targeted test scope for the situation.
```
theengineear: Of note — when writing a skill, I find it’s important to provide information, but not be _too_ overbearing. I have seen poor results when skills are written too rigidly. In this case, language like “use your judgement” or “choose based on context” are nice ways to ensure that the agent understands what’s possible, but still works to decide _what to do_.

.claude/skills/test/SKILL.md
```diff
@@ -0,0 +1,48 @@
...
+Preserve it or summarize plainly ("5/5 passed"). No emojis or checkmarks.
+
+## CLI
+
+Running `npm test` invokes `@netflix/x-test-cli` via `test.sh`.
+See `cd ui && npm test -- --help` for flags (URL, filtering, client). Examples:
+
+```sh
+npm test -- --url=http://localhost:8080/test/test-parser.html
+npm test -- --name-pattern=parser
```
theengineear, line 27: Stuff like this is _particularly_ helpful… otherwise agents default to `grep`-ing output and missing critical information because they are trying to limit the text they ingest from running things. It also helps hoomans follow along on what the agent is up to in a less cryptic way.

PR #361, opened 2026-03-26
Fix “noImplicitAny” errors in “x-element.js”.

theengineear:
With the change from TypeScript “5.x” to “6.x”, the “strict” config setting now defaults to “true”. We are simply working through the files one-by-one to be able to accept that new default change. It’s ultimately what we want, anyhow.

Note, this also removes the “noImplicitAny” entirely from our “tsconfig” file.


tsconfig.json
```diff
@@ -21,7 +21,5 @@
     "declarationMap": true,
     "noEmitOnError": true,
     "module": "NodeNext", // Approximates browser target.
-    // TODO: #357: Fix implicit any types.
-    "noImplicitAny": false
```
theengineear, line 25: This was the main goal. I want to take on the TS defaults. We now explicitly call things `any` — which I think is a _lesser_ evil and something we can decide to fix in the future (or not).

x-element.js
```diff
@@ -822,7 +909,13 @@ export default class XElement extends HTMLElement {
     }
   }
 
-  // Called during host initialization.
+  /**
+   * Called during host initialization.
+   * @param {any} host
+   * @param {any} property
+   * @param {any} hostInfo
```
theengineear, line 916: ^^ I plan to circle back to this in future PRs. It would be a _mess_ to try and fix these `any` types here. I suspect that we will _always_ have a few `any` types in this file — i.e., being totally dogmatic about this probably doesn’t serve us. That said, I genuinely find it useful to type out some of the internals, mostly because it improves agentic understanding of this stuff.

`any`ways… “yes”, I understand that some of these annotations look a bit silly in the interim.

theengineear:
Alright @klebba — last one in this set. I think the main downside of these changes is _bloat_. We are certainly adding a bunch of commentary / annotation lines. I am down to talk pros / cons there in a post-review setting. I am going to get this merged (but will not yet publish). Maybe we can catch up at some point.

The main upside (the one I see at least) is improved guardrails for agentic flows. I find that by taking the type to annotate the _shapes_ of things, I get better results, even if this stuff is all in the internal / private interfaces.

PR #369, opened 2026-04-24
Update “x-test” / “x-test-cli”.

theengineear:
Adopts the latest interface changes which align more-closely to Node’s `node --test`, `node:test`, and `node:assert` standards.

Some notes:

- Coverage moved to the CLI library, see `x-test.config.js`.
- Conventional `./coverage/lcov.info` files get generated when you test.
- Failures get re-iterated in output (for easy AI `tail`-ing).
- Output TAP stream is stylized in supported terminals / environments.
- Top-level interface is `load`, `suite`, `test`, and `assert` now.

Screenshot 2026-04-24 at 2 43 25 PM

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,8 +10,8 @@
       "license": "Apache-2.0",
       "devDependencies": {
         "@netflix/eslint-config": "3.0.0",
-        "@netflix/x-test": "2.0.0-rc.6",
-        "@netflix/x-test-cli": "1.0.0-rc.2",
+        "@netflix/x-test": "2.0.0",
+        "@netflix/x-test-cli": "1.0.0",
         "esbuild": "0.27.4",
         "eslint": "10.1.0",
         "eslint-plugin-jsdoc": "62.8.0",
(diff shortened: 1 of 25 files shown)


test/index.js
```diff
@@ -1,32 +1,22 @@
...
 
-// We import these here so we can see code coverage.
-import '../x-element.js';
-import '../x-parser.js';
-import '../x-template.js';
-
-// Set a high bar for code coverage!
-coverage(new URL('../x-element.js', import.meta.url).href, 100);
-coverage(new URL('../x-parser.js', import.meta.url).href, 100);
-coverage(new URL('../x-template.js', import.meta.url).href, 100);
```
theengineear, line 11: ☝️ — All coverage stuff is moved to the CLI.

test/index.js
```diff
@@ -1,32 +1,22 @@
-import { test, coverage } from '@netflix/x-test/x-test.js';
+import { load } from '@netflix/x-test/x-test.js';
```
theengineear, line 1: Top-level `test` renamed to `load`.

test/test-template-engine.js
```diff
@@ -1,4 +1,4 @@
-import { assert, describe, it } from '@netflix/x-test/x-test.js';
+import { assert, suite, test } from '@netflix/x-test/x-test.js';
```
theengineear, line 1: Top-level `describe` renamed to `suite` and top-level `it` renamed to `test` (matches classic unit testing / TAP-y language).

test/test-observed-properties.js
```diff
@@ -101,77 +101,51 @@ class TestElement extends XElement {
...
   const el = document.createElement('test-element');
   document.body.append(el);
-  assert(
-    deepEqual(el.changes, [
-      { property: 'c', value: 'undefined undefined', oldValue: undefined },
-    ]),
-    'initialized as expected'
-  );
+
+  assert.deepEqual(el.changes, [
```
theengineear, line 108: Drafting off `node:assert/strict` — `x-test` now has `assert.deepEqual` baked in. Currently, it is _only_ `assert` / `assert.deepEqual` / `assert.throws` / `assert.rejects` (until there is actual need for more).

test/test-parser.js
```diff
@@ -107,14 +92,14 @@ const html = strings => wrapper(strings);
...
 const htmlol = strings => wrapper(strings);
 
-describe('basic', () => {
-  it('empty template works', () => {
+suite('basic', () => {
+  test('empty template works', () => {
     const expectedTokens = [];
     const tokens = html``;
-    assert(deepEqual(tokens, expectedTokens), stringifyTokens(tokens));
+    assert.deepEqual([...tokens], expectedTokens, stringifyTokens(tokens));
```
theengineear, line 83: Look at that! Moving to a proper `deepEqual` implementation actually forced us to fix our tests. We were _accidentally_ failing to acknowledge that there were symbol keys on this tokens object. Now, you _have_ to deal with it in the actual `test` block. This passed before because we happened to not check non-enumerable keys in our lazy `deepEqual` implementation. I like that it forces us to `[...]`-spread here since it keys you into additional context.

package-lock.json
```diff
@@ -2678,35 +2677,6 @@
         "node": ">=8"
       }
     },
-    "node_modules/tap-parser": {
```
theengineear, line 2681: Underlying `x-test-cli` internalized its TAP interpreter. That means we got to drop a few dev lines from our `package-lock.json` file as well ❤️

theengineear:
@klebba — No rush on this, just plumbing through changes to get some more feedback on the `x-test` changes. I left commentary on the PR and I think the screen grab in the description is a good summary of what’s new.

x-test.config.js
```diff
@@ -0,0 +1,11 @@
+export default {
+  url:      'http://127.0.0.1:8080/test/',
+  client:   'puppeteer',
+  browser:  'chromium',
+  coverage: true,
+  coverageGoals: {
```
theengineear, line 6: Here’s where we configure coverage goals.

test/test-parser.js
```diff
@@ -694,78 +663,78 @@ describe('odds and ends', () => {
...
+    assert.throws(callback, new RegExp('^' + RegExp.escape(expectedMessage)));
   });
 
-  it('throws when attempting non-trivial interpolation of a textarea tag (succeeding space)', () => {
+  test('throws when attempting non-trivial interpolation of a textarea tag (succeeding space)', () => {
     const callback = () => html`<textarea id="target">${VALUE} </textarea>`;
-    const expectedMessage = '[#155]';
-    assertThrows(callback, expectedMessage, { startsWith: true });
+    const expectedMessage = 'Error: [#155]';
+    assert.throws(callback, new RegExp('^' + RegExp.escape(expectedMessage)));
```
theengineear, line 686: The `assert.throws` is also just available from `x-test` now. Again, this is copying the interface of `node:assert`. We _could_ expand this to support more than just the _pattern_ in the future. Honestly, with the addition of `RegExp.escape` — it’s not too bad to reliably convert a string to a pattern like this.

PR #370, opened 2026-05-16
ci: declare workflow-level `contents: read` on 2 workflows

arpitjain099:
Pins the default `GITHUB_TOKEN` to `contents: read` on 2 workflows in `.github/workflows/` that don't call a GitHub API beyond the initial checkout.

Left implicit because they reference `GITHUB_TOKEN` / use a write-scope action / trigger on `pull_request_target`. Best declared by a maintainer: `test.yaml`.

## Why

CVE-2025-30066 (March 2025 `tj-actions/changed-files` supply-chain compromise) exfiltrated `GITHUB_TOKEN` from workflow logs. Pinning per workflow caps runtime authority irrespective of the repo or org default, gives drift protection if the default ever widens, and is credited per-file by the OpenSSF Scorecard `Token-Permissions` check.

YAML validated locally with `yaml.safe_load` on each touched file.

### DIFF
--- a/.github/workflows/check-main.yaml
+++ b/.github/workflows/check-main.yaml
@@ -3,6 +3,10 @@ name: Check Main
 on:
   push:
     branches: [main]
+
+permissions:
+  contents: read
+
 jobs:
   test:
     uses: ./.github/workflows/test.yaml
--- a/.github/workflows/check-pull-request.yaml
+++ b/.github/workflows/check-pull-request.yaml
@@ -2,6 +2,10 @@
 name: Check Pull Request
 on:
   pull_request
+
+permissions:
+  contents: read
+
 jobs:
   test:
     uses: ./.github/workflows/test.yaml


klebba:
Hi @arpitjain099 — can you help me to understand how our workflows are impacted? I do not follow how our workflows have any relationship with [tj-actions/changed-files](https://github.com/tj-actions/changed-files)

arpitjain099:
Hey @klebba, fair question. You're right, you're not using `tj-actions/changed-files` here, so this PR isn't a fix for that specific bug.

I keep citing tj-actions because it's the loudest recent example, but the underlying issue is generic. When any third-party GitHub action gets supply-chain compromised, the damage to each caller equals whatever scope `GITHUB_TOKEN` was issued with at the workflow level. The malicious tj-actions version dumped that token into workflow logs. Repos that had already scoped their token to `contents: read` leaked something basically harmless. Repos still on the default "read and write" repo token leaked something that could push, edit PRs, edit releases.

So the bet this PR is making isn't "you have a vulnerable action today". It's: if `actions/checkout`, `actions/setup-node`, or anything in the reusable `test.yaml` chain gets compromised tomorrow, the token leaked from your workflow would be `contents: read` only rather than full repo write. The two workflows in this PR just trigger downstream tests (npm ci, lint, type, build, test), and nothing in that chain needs the token to do anything past the initial checkout, so dropping to `contents: read` is a no-op functionally.

OpenSSF Scorecard's `Token-Permissions` check also grades this per-file fwiw, so explicit declarations bump that score by one per workflow.

If Netflix's stance is "we don't pin scopes preemptively on workflows that aren't currently broken", totally fine to close, no hard feelings.

klebba:
Reported: `Please select a category for your issue: Spam or inauthentic Activity`

arpitjain099:
Closing, sorry for the noise.

PR #374, opened 2026-08-15
Fix reference decoding in attribute values.

### LINKED ISSUE #373: Attribute values and textarea content lose markup-shaped text
> Attribute values and textarea content lose markup-shaped text when they also contain a character reference

Claude shared a quibble that lead to a potential simplification and bug fix. Could be worth a look @theengineear 

---

Attribute values lose markup-shaped text when the value also contains a character reference

### Repro

Reading `getAttribute('title')` after each render:

```js
html`<div title="<placeholder>"></div>`            // "<placeholder>"          ok
html`<div title="</template>"></div>`              // "</template>"            ok
html`<div title="a < b &amp; c"></div>`            // "a < b & c"              ok

html`<div title="<placeholder> &amp; more"></div>` // " & more"                want "<placeholder> & more"
html`<div title="<img src=x> &amp;"></div>`        // " &"                     want "<img src=x> &"
html`<div title="x</template>y &amp; z"></div>`    // "xy & z"                 want "x</template>y & z"

html`<textarea><em>this</em> plain</textarea>`
// "<em>this</em> plain"                        ok

html`<textarea><em>this</em> is the &ldquo;default&rdquo; value</textarea>`
// "this is the “default” value"                want "<em>this</em> is the “default” value"
```

Wanted values match native parsing of the same markup.

Two conditions have to coincide: the value contains text the html parser will consume as markup — a tag, end tag, or comment — **and** it contains at least one character reference. Either alone is fine, which is why the first three lines pass.

### Cause

`#onToken` accumulates plaintext and reference tokens into `state.text` — `attributeValuePlaintext`/`attributeValueReference` for attribute values, `textPlaintext`/`textReference` for text and textarea content — then decodes the whole run at `attributeValueEnd` or `textEnd`. `#decode` uses `setHTMLUnsafe`, so once a reference has put the run on the decode path, plaintext alongside it is re-parsed as markup — `` becomes an element and drops out of `textContent`. Without a reference, `state.encoded` stays false, nothing is re-parsed, and the value survives.

An attribute value is not markup. Per the tokenizer's *attribute value (double-quoted) state*, `&` diverts to the character reference state and everything else, `<` included, is appended literally.

Ordinary text runs are unaffected: `XParser.#text` is `/[^<]+/y` and rejects `<` outright. Textarea content is not an ordinary text run — it is replaceable character data, matched by `#throughTextarea` (`/.*?<\/textarea>/ys`), which permits `<` and emits it as `textPlaintext`:

plaintext("<em>this</em> is the ") + reference("&ldquo;") + plaintext("default") + reference("&rdquo;") + plaintext(" value")

So `<` reaches the decoder from two directions — attribute values and textarea content — and all three contexts share the same cases in `#onToken`. That sharing is why it was missed.

`test-parser.js:668` already contains the textarea string above. It passes because that suite asserts the token stream and never renders.

### Fix

Decode at the reference cases instead of the end cases, so `#decode` only ever receives one entity. `#decode` and `#htmlEntityContainer` are unchanged — `setHTMLUnsafe` still performs every decode.

```diff
       case XParser.tokenTypes.textReference:
       case XParser.tokenTypes.attributeValueReference:
-        state.text += substring;
-        state.encoded = true;
+        state.text += TemplateEngine.#decode(substring);
         break;
       case XParser.tokenTypes.textEnd: {
-        const decoded = state.encoded ? TemplateEngine.#decode(state.text) : state.text;
         if (
           state.tagName === 'pre' &&
           state.childNodesIndex === -1 &&
-          decoded.startsWith('\n')
+          state.text.startsWith('\n')
         ) {
-          state.element.appendChild(TemplateEngine.#document.createTextNode(decoded.slice(1)));
+          state.element.appendChild(TemplateEngine.#document.createTextNode(state.text.slice(1)));
         } else {
-          state.element.appendChild(TemplateEngine.#document.createTextNode(decoded));
+          state.element.appendChild(TemplateEngine.#document.createTextNode(state.text));
         }
         state.childNodesIndex += 1;
-        state.encoded = false;
         state.text = '';
         break;
       }
       case XParser.tokenTypes.attributeValueEnd: {
-        const decoded = state.encoded ? TemplateEngine.#decode(state.text) : state.text;
-        state.element.setAttribute(state.name, decoded);
+        state.element.setAttribute(state.name, state.text);
         state.name = '';
-        state.encoded = false;
         state.text = '';
@@
         childNodesIndex: -1,
-        encoded: false,
         text: '',
         name: '',
```

That removes every `state.encoded` site: the write at the reference case, both reads (each embedded in a `const decoded = …` line), both resets, and the field itself.

### Cost

One `#decode` call per reference instead of one per run, at analysis time behind `#stringsToAnalysis`:

| run | current | proposed |
| --- | --- | --- |
| one reference | 0.0020 ms | 0.0020 ms |
| eleven references | 0.0030 ms | 0.0207 ms |

Eleven-reference case taken from `renders references for commonly-used characters`.

Suite stays green with the diff applied.

theengineear:
Previously, we accumulated text that had encoded entities (e.g., like “&amp;”) until we hit the _end_ of the text — then decoded the whole string at once via `setHTMLUnsafe`. BUT, this wrongly was parsing values with _tags_ into HTML (when all we wanted was to decode entities).

The fix is to _only_ use `setHTMLUnsafe` to decode exactly one HTML entity at a time. Theoretically, this could cause performance issues for some edge / pathological cases.

Note: this also fixes text parsing / decoding for `<textarea>` content which suffered from the same bug.

Closes #373.

### DIFF
--- a/test/test-template-engine.js
+++ b/test/test-template-engine.js
@@ -84,6 +84,12 @@ suite('html rendering', () => {
     assert(container.children[0].textContent === `--&:^);--`);
   });
 
+  test('renders non-interpolated <textarea> content with special characters and html entities', () => {
+    const container = document.createElement('div');
+    render(container, html`<textarea><em>this</em> is the &ldquo;default&rdquo; value</textarea>`);
+    assert(container.querySelector('textarea').value === '<em>this</em> is the “default” value');
+  });
+
   test('renders interpolated content without parsing', () => {
     const userContent = '<a href="https://evil.com">Click Me!</a>';
     const container = document.createElement('div');
@@ -254,6 +260,12 @@ suite('html rendering', () => {
     render(container, html`<div foo="--&#123;&lt;&amp;&gt;&apos;&quot;&#x007D;--"></div>`);
     assert(container.childElementCount === 1);
     assert(container.children[0].getAttribute('foo') === `--{<&>'"}--`);
+  });
+
+  test('renders non-interpolated attribute values with special characters and html entities', () => {
+    const container = document.createElement('div');
+    render(container, html`<div title="<code>tags</code> &amp; more"></div>`);
+    assert(container.children[0].getAttribute('title') === '<code>tags</code> & more');
   });
 
   test('renders boolean attributes', () => {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   `useUnknownInCatchVariables`, and `strictFunctionTypes` (#357).
 - Bump devDependencies (eslint 10, typescript 6, eslint-plugin-jsdoc 62, etc.).
 
+### Fixed
+- Markup-shaped plaintext (e.g., `<code>`) is no longer lost from attribute
+  values and `textarea` content when it appears alongside a character reference
+  (e.g., `&amp;`). Previously, the presence of any reference caused text to be
+  re-parsed as markup (#373).
+
 ## [2.0.2] - 2026-03-10
 
 ### Fixed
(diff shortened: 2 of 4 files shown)


test/test-template-engine.js
```diff
@@ -256,6 +262,12 @@ suite('html rendering', () => {
     assert(container.children[0].getAttribute('foo') === `--{<&>'"}--`);
   });
 
+  test('renders non-interpolated attribute values with special characters and html entities', () => {
+    const container = document.createElement('div');
+    render(container, html`<div title="<code>tags</code> &amp; more"></div>`);
```
theengineear, line 267: I think this test really covers what we want to test (without being overly verbose). There is one for attribute values and one for `<textarea>` content.

